PR #3, opened 2018-03-06
Minor changes to templates

kafitzgerald:
Minor changes to templates for consistency w/ v5.0.0 filenames and more relevant comment in setEnvar.sh.


kafitzgerald:
I had forgotten that this was never merged.  Think I was waiting for comments.  

These should be pretty minor changes (only to templates), but could be problematic if testing still relies on these.  @jmills-ncar do you have a quick answer on this?

PR #10, opened 2018-03-23
CHRTOUT output routine fix for io_form_outputs=0 and SOILM check tolerance error fix

aubreyd:
Two small housekeeping fixes (should not affect answers):
1) Switched the output option CHRTOUT_DOMAIN=1 when io_config_outputs=0 to call output_chrt2 since output_chrt is broken (and deprecated twice over). Only needed for Noah runs.
2) Very small (<10^-8) SMC>SMCMAX diffs causing fatal error in aggregation in some calib runs. Changed to very small tolerance check.

### DIFF
--- a/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
+++ b/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
@@ -372,34 +372,35 @@ if(nlst_rt(did)%SUBRTSWCRT         .gt. 0 .or. &
       endif
    else
       ! Call traditional output routines
-             if(nlst_rt(did)%CHRTOUT_DOMAIN  .eq. 1)  then
-#ifdef MPP_LAND
-                 call mpp_output_chrt( &
-                      rt_domain(did)%gnlinks,rt_domain(did)%gnlinksl,rt_domain(did)%map_l2g, &
-#else
-                 call output_chrt(  &
-#endif
-                   nlst_rt(did)%igrid, nlst_rt(did)%split_output_count, &
-                   RT_DOMAIN(did)%NLINKS,RT_DOMAIN(did)%ORDER, &
-                   nlst_rt(did)%sincedate,nlst_rt(did)%olddate,RT_DOMAIN(did)%CHLON,&
-                   RT_DOMAIN(did)%CHLAT,                                      &
-                   RT_DOMAIN(did)%HLINK, RT_DOMAIN(did)%ZELEV,                &
-                   !RT_DOMAIN(did)%QLINK,nlst_rt(did)%DT,Kt,                   &
-                   str_out, nlst_rt(did)%DT,Kt,                   &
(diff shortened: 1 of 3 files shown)


trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
```diff
@@ -372,34 +372,35 @@ subroutine HYDRO_out(did, rstflag)
...
-                   , RT_DOMAIN(did)%nudge                                     &
-#endif
-                   , RT_DOMAIN(did)%accSfcLatRunoff, RT_DOMAIN(did)%accBucket &
-                   , RT_DOMAIN(did)%qSfcLatRunoff,   RT_DOMAIN(did)%qBucket   &
-                   , RT_DOMAIN(did)%qin_gwsubbas                              &
-                   , nlst_rt(did)%UDMP_OPT                                    &
-                   )
-              else
-                if(nlst_rt(did)%CHRTOUT_DOMAIN  .eq. 2)  then
+!ADCHANGE: We suspect this routine is broken so default is now output_chrtout2
```
kafitzgerald: Rather than comment, should we just delete?

trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
```diff
@@ -1152,9 +1153,9 @@ subroutine aggregateDomain(did)
                    endif
 
 !end 
-                    IF (RT_DOMAIN(did)%SMCRT(IXXRT,JYYRT,KRT) .GT. &
-                        RT_DOMAIN(did)%SMCMAXRT(IXXRT,JYYRT,KRT)) THEN
 
+!ADCHANGE: Change to accept very small diff since was being triggered by vals that print identical
+IF ( (RT_DOMAIN(did)%SMCRT(IXXRT,JYYRT,KRT)-RT_DOMAIN(did)%SMCMAXRT(IXXRT,JYYRT,KRT)) .GT. 0.000001 ) THEN
```
kafitzgerald: This seems to work well (solved some issues I was having), but I wonder about the hard coded value.  

Seems like there are maybe [better practices](http://jules-lsm.github.io/coding_standards/guidelines/fp_arithmetic.html).  Also, for reference, Noah / Noah deal w/ this by defining a constant [here](https://github.com/NCAR/wrf_hydro_nwm_public/blob/master/trunk/NDHMS/Land_models/Noah/Utility_routines/module_model_constants.F).

jmccreight:
I totally agree with both of these. I'm personally suffering confusion since yesterday because of my own use of a magic/hard-coded number.

jmccreight:
@aubreyd What do you think?

jmccreight:
Maybe we should start a constants module?

jmccreight:
I opened an issue on a constants module: see #28 
@aubreyd, what do you think about deleting the code that was commented instead of commenting it?

PR #33, opened 2018-04-26
Updates to tests and json files

jmills-ncar:


jmills-ncar:
OK, testing fails because the reference (NCAR/master) does not contain the testing updates. However, the candidate (this pr) passes relevant tests so this is OK to merge from that perspective

jmccreight:
Expected testig failure...

PR #39, opened 2018-05-04
Logans phone number in there twice.

jmccreight:


jmccreight:
This test is failing very unexpectedly. I'm going to have to take a look and this may sit over the weekend.

kafitzgerald:
The fail isn't just the metadata issue?

jmccreight:
I do NOT get this error locally.... hmmm

jmccreight:
We determined that @jmills-ncar is an xarry n00b who didnt notice that the output types were not the required types for RouteLink. We also logged #40 to trap these at run time.

PR #41, opened 2018-05-08
Metadata fix

arezoorn:
This has three of the requested metadata by the NWC in it. 

1. two minor spelling issues

2. adding the Model output type (model_output_type=“forcing”, “land”, “reservoir”, “channel_rt”, “terrain_rt”), i have added one for groundwater also called "groundWater_rt", this was not on NWC list, since ground water is not outputted operationally. 

3. adding Model configuration type (model_configuration=“analysis_and_assimilation”, “short_range”, “medium_range”, “long_range”). Our list has three more options, I have used the following as the model_configuration names. Suggest other names if you do not find the suggested ones good. 
(0 = “default”
1 = "analysis_and_assimilation"
2 = "short_range"
3 = "medium_range"
4 = "long_range"
5 = "retrospective"
6 = "diagnostic")


trunk/NDHMS/Routing/module_NWM_io_dict.F
```diff
@@ -510,6 +530,9 @@ module module_NWM_io_dict
    character (len=64), dimension(numGwVars) :: varNames
    integer :: numVars = numGwVars
    character (len=512) :: outVersion = codeVersion
+   character (len=512) :: modelOutputType = "groundWater_rt"
```
kafitzgerald: Super minor thing here, but my vote would be for "groundwater_rt" since USGS + the scientific community made a rather large point of saying that groundwater should now be one word back in 2009 or so.  Memos for reference, context, and humor:

https://water.usgs.gov/admin/memo/GW/gw75.03.html
https://water.usgs.gov/admin/memo/GW/gw09.03.html

And in the journal Groundwater (though it may have been Ground Water at the time ;) ):
https://doi.org/10.1111/j.1745-6584.2009.00566.x

arezoorn: Good comment! I did not know that. I will fix this.

kafitzgerald:
@logankarsten approved the updated PR on the private side so I'm going to go ahead and merge.

PR #43, opened 2018-05-10
Metadata fix

arezoorn:
I have added the min_valid and max_valid as attributes to the output files from NWM output routines. To do so, I added KHOUR to the nlst_rt, since the simulation duration was not accessible in other routines. 

Potential problem: I noticed that in the NWM output routine, the simulation time is calculated based on the out_dt for all the routines, even for output_NoahMP_NWM and output_lsmOut_NWM subroutines. Should not we be using the dt (NOAHMP time step) for these ones? For the NWM configuration might not be important since the time step for outputting is the same between the LSM and Hydro, but we will have issues if not. If someone could confirm my suspicious, then I will fix them accordingly.


kafitzgerald:
Closing for now.

PR #45, opened 2018-05-12
Metadata fix

arezoorn:
I have added the min_valid and max_valid as attributes to the output files from NWM output routines. To do so, I added KHOUR to the nlst_rt, since the simulation duration was not accessible in other routines.

I also added one global attribute called model_total_valid_times which is basically the total number of outputs for a given file format (for example CHRTOUT) during the simulation.

I also resolved NCAR/wrf_hydro_nwm-issues#139 . The NoahMP time stamp was wrong (when the time stamp between LSM and Hydro differed). I added ITIME which is the LSM time step to the output routine and used that instead of out_counts which is Hydro time step. I also ran the model with LSM only, checked that the time is correct. 

There was also a non reported issue/bug that is fixed with this change, the time variable was wrong if HYDRO was active and the LSM output timestep was not a multiplicative of the Hydro output timestep. For example if user was setting the LSM output timestep to 36000 (10hours) and Hydro output timestep to 180 (3 hours) the time variable in the LSM output was wrong. This was not a reported bug, since in our operational runs and our tests, we usually define the LSM output timestep a miltuplicative factor of the LSM output.


kafitzgerald:
Note that if this review needs to happen on the private repo we need to clear or close the previous PR first.  Just let me (or James or Joe) know if that needs to happen

arezoorn:
I am no sure if I understand your concern. Eventually it goes to private, but I guess the plan is to do it first n the public side. I do not know what is the process, and leave it to you, James or Joe to decide what needs to happen first.

jmccreight:
There is no comment that the PR passes any kind of regression test. 
No test results have been placed in-line. 
Can someone please comment on why the checks are failing? If the report is hard to read, then state that. We can make it better. Right now this is completely being ignored. To me, it looks like it's probably all metadata differences but I think we need to clearly separate metadata and value differences in the evaulation and I would open an issue about that.

jmccreight:
Given that the restart files all check and the CHRTOUT and CHANOBS files fail, I'm going to assume those differences are meta-data only. 
I filed #50 to request separation of data and metadata checks for output files.

trunk/NDHMS/Land_models/NoahMP/IO_code/module_NoahMP_hrldas_driver.F
```diff
@@ -1518,7 +1523,7 @@ subroutine land_driver_ini(NTIME_out,wrfits,wrfite,wrfjts,wrfjte)
        print*, "t0OutputFlag: ", t0OutputFlag  
 #endif        
 !#ifdef HYDRO_REALTIME
-   if(t0OutputFlag .eq. 1) call ldas_output()
+   if(t0OutputFlag .eq. 1) call ldas_output(0)
```
jmccreight: I love how this is so opaque. What was the assumed value when nothing was passed, what does 0 represent. Is there no good variable that can be used?

trunk/NDHMS/Land_models/NoahMP/IO_code/module_NoahMP_hrldas_driver.F
```diff
@@ -1837,7 +1842,9 @@ subroutine land_driver_exe(itime)
 end subroutine land_driver_exe
 
 !!===============================================================================
-subroutine  ldas_output()
+subroutine  ldas_output(itime)
+integer, intent(in)  :: itime ! time step of the LSM, had to explicitly declare it since 
+                              ! it did not exist in the land_driver_ini
```
jmccreight: 👍 The comment is good, but notice how this kind of comment goes stale "had to explicitly..."
I love keywords. I wish they were slightly easier to use in fortran.

trunk/NDHMS/Routing/module_NWM_io.F
```diff
@@ -514,6 +514,13 @@ subroutine output_chrt_NWM(domainId)
    fileMeta%initTime = trim(initTime)
    fileMeta%validTime = trim(validTime)
  
+   ! calculate the minimum and maximum time
+   fileMeta%timeValidMin = minSinceEpoch1 + nlst_rt(1)%out_dt 
+   fileMeta%timeValidMax = minSinceEpoch1 + int(nlst_rt(1)%khour * 60/nlst_rt(1)%out_dt) * nlst_rt(1)%out_dt
+
+   ! calculate total_valid_time
+   fileMeta%totalValidTime = int(nlst_rt(1)%khour * 60 / nlst_rt(1)%out_dt)  ! # number of valid time (#of output files)
+
```
jmccreight: I get why these time calculations are here. I would just like to point out that these kinds of calculations should go in a module_time.F and the results should be requested here (not calculated).

jmccreight: #51

jmccreight:
This PR needs a bit more explicit description of what the changes were. @arezoorn please correct if any of this is wrong:

"I have added the min_valid and max_valid as attributes to the output files from NWM output routines. " Specifically, in module_NWM_io_dict:
```
integer :: timeValidMin ! the minimum time each configuration can have, time of the first output file
integer :: timeValidMax ! the maximum time each configuration can have, time of the last output file
integer :: totalValidTime ! # number of valid time (#of output files)
```
These are now metadata attributes in the CHANOBS and CHRTOUT (same in both) with the following names:
```
iret = nf90_put_att(ftn,timeId,'valid_min',fileMeta%timeValidMin)
iret = nf90_put_att(ftn,timeId,'valid_max',fileMeta%timeValidMax)
iret = nf90_put_att(ftn,NF90_GLOBAL,"model_total_valid_times",fileMeta%totalValidTime)
```

kafitzgerald:
Definitely agree that it would be nice to have some additional info regarding test failures in the log.

Just ran the tests again offline to confirm (file diffs below) and as expected we just see differences in the attributes.

CHRTOUT
```DIFFER : NUMBER OF GLOBAL ATTRIBUTES : 16 <> 17
DIFFER : NAME OF GLOBAL ATTRIBUTE : model_total_valid_times : GLOBAL ATTRIBUTE DOESN'T EXIST IN NWM/run_reference/201108281400.CHRTOUT_DOMAIN1
DIFFER : VARIABLE "time" IS MISSING ATTRIBUTE WITH NAME "valid_min" IN FILE "NWM/run_reference/201108281400.CHRTOUT_DOMAIN1"
DIFFER : VARIABLE "time" IS MISSING ATTRIBUTE WITH NAME "valid_max" IN FILE "NWM/run_reference/201108281400.CHRTOUT_DOMAIN1"
DIFFER : NUMBER OF ATTRIBUTES : VARIABLE : time : 3 <> 5```

CHANOBS
```DIFFER : NUMBER OF GLOBAL ATTRIBUTES : 16 <> 17
DIFFER : NAME OF GLOBAL ATTRIBUTE : model_total_valid_times : GLOBAL ATTRIBUTE DOESN'T EXIST IN NWM/run_reference/201108281400.CHANOBS_DOMAIN1
DIFFER : VARIABLE "time" IS MISSING ATTRIBUTE WITH NAME "valid_min" IN FILE "NWM/run_reference/201108281400.CHANOBS_DOMAIN1"
DIFFER : VARIABLE "time" IS MISSING ATTRIBUTE WITH NAME "valid_max" IN FILE "NWM/run_reference/201108281400.CHANOBS_DOMAIN1"
DIFFER : NUMBER OF ATTRIBUTES : VARIABLE : time : 3 <> 5```

trunk/NDHMS/Land_models/NoahMP/IO_code/module_NoahMP_hrldas_driver.F
arezoorn: hmm, that is the initialization timestep, actually I could use ITIME as I used it in the other routines. I will check why I did not use it, and it possible I would replace it with ITIME.

PR #47, opened 2018-05-15
Replicate SM checks to happen after hydro2dtbl read for consistency.

aubreyd:
Addresses Issue #127 in wrf_hydro_nwm ("Soil properties checks if reading from HYDRO_2D_TBL"). Ran the following checks using Croton on hydro-c1 PGI:
1) Confirmed checks applied as expected.
2) Confirmed TBL and NC file runs give identical solutions.
3) Confirmed same answers if urban soil properties specified in hydro2dtbl input file or hydro2dtbl is internally-generated from TBL.
4)  Answer changes from master code simulation if urban cells present and urban properties NOT specified in hydro2dtbl. This is expected behavior per Issue #127.

Also confirmed behavior on Cheyenne Intel using NWM CONUS domain. Expected differences in urban cells. Passed perfect restart with NWM CONUS.

NOTE: Per Issue #127 exchange, this is a consistency fix between LSM and hydro for hard-coding of urban soil properties. Both will be relaxed in future versions.

### DIFF
--- a/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
+++ b/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
@@ -1750,6 +1750,15 @@ end subroutine HYDRO_ini
        ! input from HYDRO.TBL.nc file
        print*, "reading from hydrotbl_f(HYDRO.TBL.nc)  file ...."
        call hdtbl_in_nc(did)
+       !ADCHANGE: For consistency, mirror urban and param value checks used in table read
+       where (rt_domain(did)%VEGTYP == rt_domain(did)%isurban) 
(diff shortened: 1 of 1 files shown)


aubreyd:
Re the restarts: No idea - should fail all if they are using the same hydro2dtbl file that has urban cells.

Re the checks: yes, these should probably be combined at some point, but this is really a temporary patch and not the way we will be doing this in the future, so I put in the minimal code changes to make it work correctly for now.

jmccreight:
@jmills the tight integration of our code and testing has revealed a mis-sepcification of the hydro tbl file in the reach based routing as of the last test.

kafitzgerald:
Ok, all three configs are failing the expected tests now.

PR #48, opened 2018-05-16
Coupled build process

kafitzgerald:
These changes address issue #166 on the private repo.  While there is still work to be done in terms of documentation and developing test cases, this fixes the build process for coupled.

This PR includes two minor changes:
* Moves `wrf_hydro_config` back into the main dir
* Fixes a reference in `module_wrf_HYDRO.F`

Testing status:
* Coupled - The coupled model now builds and runs w/ an older coupled test case, but has not been thoroughly tested. 
* Standalone - Have confirmed in offline as well as CI tests that as expected these changes do not affect the standalone configuration of the model.


kafitzgerald:
Thanks @barlage!  Good point.

PR #53, opened 2018-05-21
Putting testing into the main repo

jmccreight:
This moves the testing to the repos. This is justified by the integration and the ease of testing.

Currently things work but polish is not quite there.

CI testing can be invoked locally (requires python 3.6.4) with the following invocation:
```
cd tests/
python take_test.py
```

Options available:
* --domain: the domain (specific domain conventions required)
* --config: keywords found in the *.json files in trunk/NDHMS, e.g. NWM, Gridded, Reach
* --candidate_spec_file: see tests/template_candidate_spec.yaml
* --test_spec: currently inactive.

Tested on 
* docker
* cheyenne
  * interactive
  * scheduled with PBS

Shortcomings currently: 
* Logging to a file is not really working
* A generic calling script not done
* Still working on building more domains 
* Too many unnecessary warnings in default setups
* Have not built a README.md for the test/ dir.

Tested on cheyenne using PBS: 
```
docker@5d5f9a2c65fd:~/wrf_hydro_nwm_myFork/tests$ python take_test.py
=================================================================
*** take_test.py: A wrf_hydro candidate takes a test. ***

=================================================================
Setup the specifictions (specs):
Establish candidate spec.
Establish user spec.
Establish machine spec.
Establish test spec.

=================================================================
Boilerplate:
Date                  : 2018 May 21 04:54:10 
User                  : docker
Machine               : 5d5f9a2c65fd
take_test.py location : /home/docker/wrf_hydro_nwm_myFork/tests
Candidate spec file   : /home/docker/wrf_hydro_nwm_myFork/tests/default_candidate_spec_docker.yaml
Machine spec file     : /home/docker/wrf_hydro_tests/machine_spec.yaml
Machine spec set by   : wrf_hydro_tests_dir
User spec file        : /home/docker/wrf_hydro_tests/.wrf_hydro_tests_user_spec.yaml
User spec set by      : env var
Test spec file        : /home/docker/wrf_hydro_tests/tests/fundamental.py
Test spec set by      : command line key
Log file              : 
Will echo specs to log at end.

=================================================================
Establish repositories:

Candidate repo

Reference repo
dir_for_clone:  /home/docker/take_test/test_repos/reference
Cloning NCAR/wrf_hydro_nwm into /home/docker/take_test/test_repos/reference using https ...
Cloning into '/home/docker/take_test/test_repos/reference'...
remote: Counting objects: 11861, done.
remote: Compressing objects: 100% (55/55), done.
[...]

PR #55, opened 2018-05-22
Snow albedo IO changes and snow model density bug fix

aubreyd:
Includes the following commits migrated from wrf_hydro_nwm repo:
- Propagate BATS snow albedo variables (VIS/NIR and direct/diffuse) from subroutine albedo to module_NoahMP_hrldas_driver (barlage)
- Add several BATS-relevant snow albedo parameters to the Noah-MP parameters data structure; commit not checked for exact reproduction of hard-coded version (barlage)
- Add new snowpack albedo vars to output routines. Temp fix for isnan on PGI (aubreyd)
- Initialize new snow albedo output to zero; also set to zero when no snow present (barlage)
- Add in new snow albedo terms (and associated vars) to retro outputs. (aubreyd)
- Bug fix: change in Noah-MP subroutine SNOWH2O to update snow liquid water as water passes to layers below; solves many spurious snow density values during melt season (barlage)
- Change limit for maximum snow water depth to 5000mm from 2000mm in Noah-MP subroutine SNOWWATER (barlage)
- Remove 800mm SWE limit in Noah-MP subroutine SNOW_AGE that resets the snow age in BATS snow albedo option (barlage)

There will be expected answer changes where snow is present due to the bug fix and limit increases. These changes have been validated for CONUS NWM as part of NWM v2.0 calibration prep. New LDASOUT output variables for "retro" and "all" options. New snow parameters added to MPTABLE.TBL (included in PR) so requires TBL update to run. Compile and run tested on Cheyenne Intel 16.0.3.


jmccreight:
This will not be merged until after the release.

jmccreight:
Ideally, this will also be merged after #53, if we can make that happen today. @jmills-ncar

PR #56, opened 2018-05-22
Overland modularization changes applied to public repository 

donaldwj:
see https://github.com/NCAR/wrf_hydro_nwm/pull/170


jmccreight:
Thanks Donald. I'm going to hold on this until after the 3 other commits (2 public and 1 private) which constitute other v2 additions. Should be later this week.

jmccreight:
Regression failures are expected. These result from changes of associativity [ (a+b) + c vs a + (b+c) ] with floating point math. This was documented in [private PR 170](https://github.com/NCAR/wrf_hydro_nwm/pull/170).

PR #72, opened 2018-06-25
cache-friendly column swap

afanfa:
The double loop to swap the columns is not cache-friendly. By swapping entire columns using the array syntax we obtain better performance.


jmccreight:
Going to make sure this works before reopening...

PR #76, opened 2018-06-28
Unit fix for soil column bottom runoff for CHRTOUT/Channel-only fluxes

jmccreight:
I think one of these was not fixed. @logankarsten to verify.


kafitzgerald:
I'm going to close this and log a more complete issue.  

The good news:
The units are correct (though arguably confusing) for NWM configurations (qBtmVertRunoff is qin_gwsubbas in the code and units are m^3 accumulated over the timestep).

The bad news:
For non-NWM (UDMAP=0) configurations the simp_gw_buck subroutine has qin_gwsubbas in units of m^3/s.

PR #83, opened 2018-07-12
Correct reference repo for testing.

jmccreight:
whoopsie. minor fix but critical for local testing.


jmccreight:
rolling this into private_master PR.

PR #91, opened 2018-07-19
Re-remove retdeprt limit

aubreyd:
Re-remove retdeprt limit that was inadvertently put back in with overland refactor. This puts the code consistent with tag 113b2badca7ec45816688f215e9d24b0098d14aa and NWMv2.0 test version. Can result in (expected) answer changes, depending on the RETDEPRTFAC values tested.


aubreyd:
Add on: with this fix and the (temporary!) revert of Donald & Nels overland routing order-of-operations fix, I get identical (B4B restart) solutions to the NWMv2 test code running CONUS over the Hurricane Irene period. Thumbs up.

PR #93, opened 2018-07-23
Cache-blocking for three-level nested loop

afanfa:
This PR completes the set of optimizations on the initialization phase.
The performance improvements brought by this PR are very evident when the model run on a small number of processes (32-64). The exact performance speedup will be reported later this week.
This PR has been tested again the current version on master for CONUS 24 hours.
The results are identical.


jmccreight:
Also confirmed bt for bit match on 24-hour CONUS run.

PR #98, opened 2018-07-26
Merge bugfix testing changes into master

jmills-ncar:
**DO NOT MERGE THIS PR WITHOUT CONFIRMATION FROM ME**

This PR will remain open while I tweak other dependent parts of the testing system. Once dependent components are updated I will confirm here and can merge


jmills-ncar:
All candidate tests pass. Reference fails because the testing upgrades are not present in reference and thus will fail

jmills-ncar:
@jmccreight If you approve we can merge this and CI testing will be upgraded on master

PR #104, opened 2018-07-30
Version Module

jmccreight:
THIS IS NOT COMPLETE. We have yet to decide on the contents of the .version and .nwm_version files.


trunk/NDHMS/.nwm_version
```diff
@@ -0,0 +1 @@
+v1.2.1+dev
```
jmccreight: this should be blank?

trunk/NDHMS/utils/Makefile
```diff
@@ -0,0 +1,35 @@
+# Makefile 
+#
+.SUFFIXES:
```
jmccreight: this lline redundant/unnecessary?

trunk/NDHMS/utils/module_version.F
```diff
@@ -0,0 +1,38 @@
...
+
+character (len=512), parameter :: wrfHydroVersion=WRF_HYDRO_VERSION
+character (len=512), parameter :: nwmVersion=NWM_VERSION
+
+private :: wrfHydroVersion, nwmVersion
+public  :: get_model_version, get_code_version, get_nwm_version`
+
+contains
+
+function get_model_version()
```
jmccreight: this function should be removed?

trunk/NDHMS/.nwm_version
kafitzgerald: yeah, probably (until we create that branch?)

trunk/NDHMS/utils/module_version.F
kafitzgerald: I think it's still useful for the "title" attribute on some output files.  I use it in module_NWM_io_dict.F.

jmccreight:
After this PR is merged, we will change the .version file and the testing to match. 
Then we will fork for V2 and give similar treatment on the branch.

PR #107, opened 2018-08-01
This PR includes a number of testing upgrades

jmills-ncar:
Only testing upgrades are included in this PR. 

**DO NOT MERGE WITHOUT JMILLS**

I need to upgrade other components of the system prior to merging


jmills-ncar:
This is OK to merge, all tests functioning again

PR #110, opened 2018-08-06
Fixed bug in local testing where candidate and reference dir copies were

jmills-ncar:
not used in the testing


jmills-ncar:
@jmccreight @kafitzgerald

jmccreight:
Gracias.

PR #115, opened 2018-08-08
Testing updates

jmills-ncar:


jmills-ncar:
Testing fails for reference runs because the nudging namelists changes have not been applied ot hte reference, they are in this PR

PR #120, opened 2018-08-09
Increment the version with the new files and metadata.

jmccreight:


jmccreight:
This has expected regression fails because of the .version file.

PR #124, opened 2018-08-09
For NWM version branches, use the TBL files for operations

jmccreight:


jmccreight:
This udates the .version file.

PR #125, opened 2018-08-09
V5.1.0

jmccreight:


jmccreight:
This updates the .version file.
I was going to wait to merge this... but forgot. It will pass when we update the testing domain version.

PR #130, opened 2018-08-14
Removed caching from pytest

jmills-ncar:
Cachingi n pytests was causing many issues on cheyenne. Thus, it has been removed from the local testing run_tests.py convenience script


jmccreight:
Looks good.

PR #131, opened 2018-08-14
Minor Version update

jmccreight:
CONUS files update,
croton testing changes


jmccreight:
Version number increment will break testing

jmccreight:
We've changed the way domain versioning is handled to not break across changes in minor version numbers.

PR #135, opened 2018-08-15
V5.1.0 to master

jmccreight:


jmccreight:
I still dont love this...

jmills-ncar:
@jmccreight can you elaborate?

jmills-ncar:
Oh, so if one were to revert and run a NWM the outputs would have the v2.0 tag when infact the yare way behind the actual released v2.0 that happens on switch to production?

jmccreight:
The point is that I would prefer if the history along master did not contain a version number in the .nwm_version file. 
Note that that version number only pripigates when run with NCEP_WCOSS=1. 
The point is that when master and v5.1.0 diverge (which has not happend yet), there will be a distinct master history with .nwm_version file that is  not accurate. I might look some more into solving this, but not like I have copious time.

PR #141, opened 2018-08-17
Add Hawaii to templates and testing

kafitzgerald:
This PR will add Hawaii namelists and testing / domain info.  

Note: DO NOT MERGE yet. 

To do:
- [x] Add files to archive location and edit script for metadata generation


jmccreight:
So the tests pass except for 

1. nans in hlink, which is fine because compound channel is off.
1. the channel-only fails to match full-model under intel (this is an outstanding bug).

I think it is mergable

jmccreight:
I think we are all good. No code is touched, only testing. Testing will need some cleanup/adaption to HI domain, but can happen later.

PR #142, opened 2018-08-17
Test all outputs

jmccreight:
@jmills


jmills-ncar:
I approve the changes but am concerned about the failing test. The changes in this PR should not be the cause of the failures, but turning on all the outputs uncovered them

jmccreight:
The associated build
https://travis-ci.org/NCAR/wrf_hydro_nwm_public/builds/417377447?utm_source=github_status&utm_medium=notification
has some issues that we want to note and keep an eye on.

(493) There are some small numerical differences between channel-only and the full-model when one is using NWM IO and the other is using full outputs. These are about e-06, but should be investigated.  (#147)

(1125) Long-range, feature_id in channel-only is different?

(1512) gridded fails restart test.

PR #144, opened 2018-08-18
Fix dates in HI namelists

jmccreight:


jmccreight:
Only testing and metadata.

PR #146, opened 2018-08-18
V5.1.0 update

jmccreight:


jmills-ncar:
# NOTE VERSION NUMBERS WILL NEED TO BE SCRUBBED IN A SEPARATE PR AFTER THIS IS MERGED

trunk/NDHMS/.nwm_version
```diff
@@ -1 +1 @@
-none
```
kafitzgerald: I get why this is here, but just making a note to change it back to none.

trunk/NDHMS/hydro_namelists.json
```diff
@@ -3,32 +3,32 @@
...
             "chanrtswcrt": 1,
             "chrtout_domain": 1,
-            "chrtout_grid": 0,
+            "chrtout_grid": 1,
             "compound_channel": true,
             "dtrt_ch": 300,
             "dtrt_ter": 10,
             "dxrt": 250,
-            "frxst_pts_out": 0,
+            "frxst_pts_out": 1,
```
kafitzgerald: Base is NWM-ish, correct?  If that's the case, I don't think we can even write forecast points output.  I could be wrong here though.

trunk/NDHMS/hydro_namelists.json
```diff
@@ -73,12 +73,9 @@
         "hydro_nlist": {
             "aggfactrt": 1,
             "dxrt": 1000,
-            "io_config_outputs": 4,
-            "io_form_outputs": 2,
             "output_channelbucket_influx": 2,
             "ovrtswcrt": 0,
             "rst_typ": 0,
-            "rtout_domain": 0,
```
kafitzgerald: Still wrapping my head around these JSON namelists for testing, but it seems like if we remove this and the default is 1 we might have a problem since long-range doesn't  do surface or subsurface routing.

jmills-ncar:
as far as I can tell this is ready to merge, just need one more approver. @kafitzgerald I believe your previous review was nulled by my changes to address your comments

Issue #152, opened 2018-08-22
Global attribute(s) in output files incorrect for sub-hourly output intervals


kafitzgerald:
## Expected Behavior
The global attribute "model_output_valid_time" should be consistent with the time variable in output files. 

The global attribute "model_initialization_time" should be correct for all possible initialization times

## Current Behavior
The minutes and seconds positions are hard-coded to 00 for both of these attributes.  

## Possible Solution
Edit these lines to allow for the flexibility to accommodate non-zero values of minutes (and seconds):
https://github.com/NCAR/wrf_hydro_nwm_public/blob/d24469af63586323f7a8d6b2359b718adb732756/trunk/NDHMS/Routing/module_NWM_io.F#L505-L514

kafitzgerald:
Addressed in merged PR #208

Issue #151, opened 2018-08-22
Allow for hydro file output intervals smaller than the LSM integration timestep


kafitzgerald:
Related to issue #150...  this is a request to remove the dependency of the hydro output file frequency on the land surface model integration time step.

Given that many of the hydro components integrate forward with a much finer time step than the land surface model (and the land surface model physics - to a degree - are dependent upon the land surface model time step), it would be nice if the output file frequency for hydro components were not tied to the land surface model integration time step.

PR #154, opened 2018-08-23
Issue 152 - Remove hard-coded values for minutes and seconds in init and valid time metadata

kafitzgerald:
This PR resolves issue #152 by assigning the minutes and seconds values for init and valid time file output metadata from the appropriate time variables rather than hard-coded 00 values for both.  

I've confirmed that the code compiles and produces appropriate file metadata for sub-hourly output intervals (as long as the LSM time step is adapted accordingly - issue #151).  Since file output for testing is hourly, we shouldn't see any changes there.  

Note:
* Not an urgent fix (that I know of) since we don't typically have sub-hourly output timesteps and do not for operational applications.


jmccreight:
again confirming that is master where you want this to go.

kafitzgerald:
I can also move this to v5.1.0 if you think that's appropriate (again, there shouldn't be any answer changes).  I just didn't want to add additional overhead there.

kafitzgerald:
Closing this and opening a new one on the correct branch.

PR #155, opened 2018-08-23
Upgrades to logging in testing

jmills-ncar:
The previous method of logging resulted in messages being hidden in the html reports. These html reports are not important because they are the only means of retrieving automated cheyenne test results.

Changes to the testing have been made to allow for both stdout logging and html logging with a command line option


jmills-ncar:
testing fails bu this is due to issues in the croton domain used for testing. The changes in this PR are only to how logging is treated in the python testing scripts and thus will have no affect on the model

PR #157, opened 2018-08-24
hydro namelist json edits

jmills-ncar:
Changes to hydronamelist to address comments on pr #146


jmills-ncar:
The checks fail perfect restarts on gridded for lakes vars in hydro restart, and for channel-only vs full model. These are known issues, merging

Issue #150, opened 2018-08-22
Add namelist checks for output timesteps


kafitzgerald:
## Expected Behavior
If an out_dt (hydro.namelist) that is incompatible with NOAH_TIMESTEP (namelist.hrldas) is specified, the model should fail with an informative error message.

## Current Behavior
Incorrect output file times are generated and hydro output files are only generated at the LSM integration time step rather than the selected output interval.

There is currently no check for this and it's not a documented limitation of the modeling system.

## Possible Solution
Add in namelist checks to ensure that out_dt is a multiple of NOAH_TIMESTEP.
Update documentation to reflect this current limitation of the model.

kafitzgerald:
Addressed in merged PR #207

PR #159, opened 2018-08-27
issue-150: add namelist check for out_dt / dt

kafitzgerald:
This minor change resolves issue #150 by adding a namelist check to ensure that out_dt in hydro.namelist is a multiple of the land surface model timestep specified in namelist.hrldas.

### DIFF
--- a/trunk/NDHMS/Data_Rec/module_namelist.F
+++ b/trunk/NDHMS/Data_Rec/module_namelist.F
@@ -190,7 +190,7 @@ CONTAINS
        endif
 #endif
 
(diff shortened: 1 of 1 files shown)


jmccreight:
can this go to v5.1.0? That is to ask, does this change anything?

kafitzgerald:
Either way is fine with me.  

It won't change any answers, just causes the model to fail w/ an appropriate error message rather than running and producing unexpected results when timesteps are not compatible.

kafitzgerald:
Closing this and opening a new one on the correct branch.

PR #164, opened 2018-08-28
Minor upgrades to testing to improve log printing

jmills-ncar:
Logs now print with better alignment to enhance readability. There are many commits here that will be squashed on merge.


jmccreight:
These will come from routine merge of v5.1.0 to master.

PR #171, opened 2018-08-30
Update HI metadata

jmccreight:


kafitzgerald:
Noting here that these updates are the result of some lake topology fixes that resulted in changes in the Route_Link.nc and LAKEPARM.nc domain files for Hawaii.

PR #182, opened 2018-09-12
Fix channel-only regression by writing forcing vars to CHRTOUT as floats

jmccreight:


jmccreight:
Something is wrong with this...

PR #183, opened 2018-09-12
Fix channel-only regression by writing forcing vars to CHRTOUT as floats

jmccreight:


jmccreight:
@jmills-ncar 
Just take a look at what changed so you're awares. You can merge or I can do it later today.

jmccreight:
This will fail regression because the precision of outputs will change against the current upstream.

jmills-ncar:
nice job 🎉

PR #185, opened 2018-09-13
Bug fix in diff printing in python test scripts and change to test duration

jmills-ncar:
Only the details for the first diff are printed in the test_fundamental ncore and perf restart tests. This has been fixed.

Additionally, @aubreyd noted that at longer timescales perfect restart tests fail for far more variables than at shorter timescales. Thus, the default duration in the .json files has been increased from 3 hours to 5 days. Indeed, the longer duration reveals far more imperfect restart issues in gridded routing.


jmccreight:
Seems legit.

PR #188, opened 2018-09-14
Some testing improvements

jmccreight:


jmccreight:
This PR passes locally. But we should scrutinize the failures here a bit before merging.

jmills-ncar:
maybe I am missing it but it appears that th output frequency is 1 hour everywhere?

jmccreight:
Yes, I set it to that since I don think that's the killer in time. That's just a change in the namelists.

jmills-ncar:
Reopening from my fork to address comments

Issue #174, opened 2018-09-06
Gridded routing with lakes fails perfect restart


aubreyd:
As shown in CI testing, Croton gridded routing configuration with lakes fails perfect restart test.

## Expected Behavior
Perfect restarts in all vars.

## Current Behavior
Per CI testing, lake_inflort does not perfect restart in Croton test case gridded routing configuration. All other vars pass, so this may be just a reporting issue (may not be impacting physics/solution).

Test log here: https://travis-ci.org/NCAR/wrf_hydro_nwm_public/builds/425358497
hydro restart diff:
Variable Group  Count   ...       Range     Mean   StdDev
0  lake_inflort     /      3   ...     3.74963  2.08033  2.13999

jmills-ncar:
I dug a little deeper on this and found the following:

1) In contrast to v5.1.0-alphaX, v5.0.2 release only fails perfect restarts for the variable `inflort` 
2) Gridded v5.0.2 passes perfect restarts when lakes are not included in the domain, indicating the problem is in code exercised by lakes.

jmills-ncar:
There also appears to be a domain issue, this could be in the files or in hte namelists, need to do more digging.

Running v5.1.0 code with v5.0.2 domain yields the same minimal fail of perfect restarts only on the `inflort` variable

Running v5.1.0 code with dev domain yields numerous fails of perfect restarts

```---------------- Starting WRF-Hydro Testing ----------------
Testing the configs: gridded


####################################
### TESTING:  ---  gridded  ---  ###
####################################

pytest -vv --tb=no --ignore=local -p no:cacheprovider  -s --ignore=tests/test_supp_1_channel_only.py  --html=/home/docker/test_out/wrfhydro_testing-gfort-gridded.html --self-contained-html --config gridded --compiler gfort --domain_dir /home/docker/test_out/example_case --candidate_dir /home/docker/test_out/candidate_can_pytest/trunk/NDHMS --reference_dir /home/docker/test_out/reference_ref_pytest/trunk/NDHMS --output_dir /home/docker/test_out --ncores 2
======================================== test session starts ========================================
platform linux -- Python 3.6.5, pytest-3.8.0, py-1.6.0, pluggy-0.7.1 -- /home/docker/miniconda3/bin/python
[...]

PR #192, opened 2018-09-18
Testing housecleaning items

jmills-ncar:
This PR brings in numerous updates to testing that were made on v5.1.0 branch. Additionally, a bug was fixed in the namelist settings where an incorrect aggfacrt and dxrt were specified in the hydro.namelist. This namelist fix relates to #174


jmills-ncar:
Regarding testing fails:

reference run will fail because the namelist changes in the PR are not in reference
gridded fails as it has for some time now

PR #193, opened 2018-09-18
Add gridded_no_lakes as a testing config

jmills-ncar:
gridded_no_lakes has been added as an optional testing configuration to the namelist.json files.

Additionally, traceback turned back off after debugging


jmills-ncar:
I also went ahead and added a fresh conda environment yml file for cheyenne testing

jmills-ncar:
Regarding testing fails

Reference run fails for nwm because the namelist changes in this PR are not in reference
Gridded fails as it has for some time now

PR #197, opened 2018-09-25
Improve the scale factor for greater accuracy to to-channel fluxes (e…

jmccreight:
…xcludes the volume from the bottom of the column used in chan-only+bucket

The imporvement results in slightly worse compression to the tune of about 2MB per chrtout file when compressed as in the operational work flow.

### DIFF
--- a/trunk/NDHMS/Routing/module_NWM_io_dict.F
+++ b/trunk/NDHMS/Routing/module_NWM_io_dict.F
@@ -732,7 +732,7 @@ subroutine initChrtDict(chrtOutDict,diagFlag,procId)
                                  "latitude longitude","latitude longitude",&
                                  "latitude longitude","latitude longitude",&
                                  "latitude longitude","latitude longitude"]
-   chrtOutDict%scaleFactor(:) = [0.01,0.01,0.1,0.01,0.01,0.001,0.001,0.001,&
+   chrtOutDict%scaleFactor(:) = [0.01,0.01,0.1,0.01,0.01,0.00001,0.00001,0.001,&
                                  0.01,0.01]
    chrtOutDict%addOffset(:) = [0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,&
                                0.0]


dblodgett-usgs:
What is the impact on in-memory / uncompressed file size? Has this change been discussed with data users and service developers? Is "accuracy" the right word here? Is the model performance good enough that this is actually storing accuracy or just creating precision?

jmccreight:
This  only affects the size of the (compressed) operational output files (and perhaps reforecast output files, if they are run through the same post processing chain as in operations when run in the reforecast). The increase in file size has a mean of 14.5% in my tests, which is about 2MB per file. This should have a minimal impact on downstream users. I've taken an in-house survey (since we pull all of these files all of the time from NOMADS). There was no real concern. The increase in this part of the data is just a small part of the total data we pull continuously from NOMADS. 

This drawback is balanced by the benefit of obtaining more accurate channel-only runs because the forcing flux terms provided in these files have increased accuracy. Yes accuracy is the correct term. There is a "true" value of the model which we are truncating when we store. Running the channel-only model with these truncated values does not accurately reproduce the full model results. With the changes in this PR the differences should be significantly reduced and hence the "accuracy" improved. 

This change was driven by the data user sof the fields that were changed (me and a substantial list others). The fields are used by select users of the model in a research mode where channel-only capability has the potential to easily save hundreds of thousands of core hours for research activities over the next 2 years or more. 

There is no impact on in-memory size. There is no impact on the uncompressed file sizes.

jmccreight:
I'll also note that there is absolutely no change in any interface to the file. Netcdf handles this change under the hood.

dblodgett-usgs:
Ahh. OK I see. Thanks for the description. 

I read this to mean it was affecting more than it sounds like it is. Se these are not data that are being pulled down to HydroShare and the NWC? Those are the kinds of users I was referring to who may care. 

We could quibble about the `"true" value of the model` being accurate or just a precise inaccurate estimate, but we'll leave that alone for now ;)

Out of curiosity, what variables is this affecting, what are their units, and what is the new valid range/precision?

jmccreight:
You can see here that
https://github.com/NCAR/wrf_hydro_nwm_public/pull/197/files
those fields are being divided by 1e-5 (i know it's cryptic, but you can see the change from 1e-3).

THe numbers have a lower bound of zero in the model. And I gathered some numbers from the NWM over the last month (ana and short range cycyes including flooding from Florence) and found that 1e5 still keeps the value under 1 billion (definitely less than 2,147,483,647). I forget but i think the maximum  flux numbers were all less than 1e3 (definitely less than 1e4).

dblodgett-usgs:
OK, Yeah, those are files that are getting replicated to [hydroshare](http://thredds.hydroshare.org/thredds/catalog/nwm/catalog.html) and the National Water Center's infrastructure, so some consideration for those downstream systems may be warranted.
 
To be clear, this change goes from:
32bit integers with a 1e3 scale factor: −2,147,483.648 to 2,147,483.647 (CMS?!?)
to
32bit integers with a 1e3 scale factor: −21,474.83648 to 21,474.83647 (CMS?!?)

OK, given that these are 32 bit integers (I was assuming 16 bit!) we are fine. However, I am not clear why the operational model needs to be producing such highly precise output for these values. This is not an issue to discuss in a PR, but the (operational) model should be reporting  precision that is commensurate with the real world accuracy of the estimates regardless of engineering requirements of model development and execution. But like I said, that's an issue for another day.

jmccreight:
Yes, those ranges are the actual ranges that netcdf could yield (i was trying to estimate the physical ranges we might see). Putting it that way indicates that we should be using an unsigned integer.... doh. Maybe next round! Yes the units are CMS.

The metadata is in the files: 
```
jamesmcc@chimayo[503]:~/Downloads> ncdump -h  nwm.t00z.analysis_assim.channel_rt.tm02.conus.nc  | egrep '(qSfcLatRunoff|qBucket)' | grep long_name
		qSfcLatRunoff:long_name = "Runoff from terrain routing" ;
		qBucket:long_name = "Flux from gw bucket" ;
```

I should mention this only affects the analysis and short-range cycles/forecasts (the smallest chrtout files of the set of cycles). And my estimates about extra bloat were basedon the analysis cycle.

The short of the answer to the philospohcal question is that it's much easier to provide a single pipeline for research and operations applications than two seperate pipelines. I mostly agree with that if it's something like this that's not too onerous and not too easily interpretable as an observable quantity in the real world! ha

PR #202, opened 2018-10-01
mostly new NWM routelink

jmccreight:


jmccreight:
There are a few nicknacks in here to observe

tests/config_file_meta/CONUS/nwm_ana/hydro.namelist
```diff
@@ -18,24 +18,24 @@
     hydrotbl_f = './NWM/DOMAIN/hydro2dtbl_FullRouting_NWMv2.0.nc'
     igrid = 1
     io_config_outputs = 0
-    io_form_outputs = 2
+    io_form_outputs = 4
```
kafitzgerald: I believe we set this to 2 in operations.  Is this just for testing?

tests/config_file_meta/CONUS/nwm_ana/hydro.namelist
```diff
@@ -18,24 +18,24 @@
...
     land_spatial_meta_flnm = './NWM/DOMAIN/GEOGRID_LDASOUT_Spatial_Metadata_1km_NWMv2.0.nc'
     lsmout_domain = 1
     nsoil = 4
     order_to_write = 1
     out_dt = 60
     outlake = 1
-    output_channelbucket_influx = 2
+    output_channelbucket_influx = 0
```
kafitzgerald: Is there a reason to turn these off?

kafitzgerald: Hawaii now appears to be inconsistent w/ CONUS

tests/config_file_meta/CONUS/nwm_ana/hydro.namelist
jmccreight: These are the testing namelists. Not operations. 
This change is required for perfect restarts on channel-only. I guess there are other sneakier ways to do this, but this is the clearest. 
We should eventually have tests for the operational outputs and such, but the current namelists focus on the engineering tests.

tests/config_file_meta/CONUS/nwm_ana/hydro.namelist
jmccreight: ... I guess we are doing the sneakier way here but not above. I will look at HI.  This is a good point.

jmccreight:
OK. I believe we should have
io_form_outputs = 4 
everywhere for sanity that we are doing engineering tests on output with full floating precision. 

I will made the 
output_channelbucket_influx = 0
set only in base, so this will turn it off for HI in testing. Channel-only tests on HI will turn it on within the test.

jmccreight:
Joe made a comment which I logged as an issue #204

PR #209, opened 2018-10-18
made exclusion variables explicit in the test

jmills-ncar:
exclusion of variables is defined in the tests instead of being excluded in wrf_hydro_py python package classes.
Added lake_inflort as excluded variable


jmills-ncar:
Closed pending new croton spinup

PR #210, opened 2018-10-18
made exclusion variables explicit in the test

jmills-ncar:
instead of excluding variables in wrf_hydro_py python package classes they are now excluded in the tests.
Added lake_inflort as excluded variable


jmills-ncar:
closed pending new croton spinup

PR #215, opened 2018-10-19
Bugfix 5.0.x

jmills-ncar:
Syncing bugfixes into master


jmills-ncar:
This needs to be closed until level pool goes in

PR #222, opened 2018-10-25
namelist.hrldas comment negative restart frequency.

### LINKED ISSUE #221: Change default value for HYDRO_D to 0
HYDRO_D=1 can crash some subprocess commands and schedulers that buffer the stdout. This can cause a silent fail of hte model run and is very confusing when it happens. 

I suggest turning HYDRO_D off by default

jmccreight:
#221 
Seems like there's other stuff in this merge that should not be in here... but maybe it should be. @jmills-ncar

### DIFF
--- a/trunk/NDHMS/template/NoahMP/namelist.hrldas
+++ b/trunk/NDHMS/template/NoahMP/namelist.hrldas
@@ -39,6 +39,7 @@
  OUTPUT_TIMESTEP  = 3600 
 
  ! Land surface model restart file write frequency
+ ! A value of -99999 will output restarts on the first day of the month only.
  RESTART_FREQUENCY_HOURS = 2
 
  ! Split output after split_output_count output times.


jmills-ncar:
Yeah looks like the merge conflicts maybe were not resolved correctly? All teh changes to test_supp_1_channel_only.py should be dropped, except the pep8 changes

jmills-ncar:
Can we close this? Seems like a duplicate of #223

jmccreight:
Are we merging bugfix to master?

jmccreight:
Also I botched this PR. I forgot it was open and force pushed to my master. 
It now has another minor, fiddly change...

barlage:
Guess I'm late to the party here, but in my understanding any restart_frequency_hours <= 0 (not just -99999) will write monthly restart files. Also, in my non-Hydro HRLDAS, I have changed this to < 0, so that = 0 will not write any restarts, but that is a separate issue.

jmccreight:
Good to know, i just tested -99999 to confirm that works. 
Do you want to put that in the comment via a new PR? Probably is not necessary at this point.

barlage:
I'd like to address this =0 issue later and can modify the comment at that time. The current comment is not wrong so I don't see any need for a new PR.

jmccreight:
I dont know how this would fit into the =0 thing, but two more related capabilities are desirable:
1) output restarts at just the final time of the run. 
2) output restarts at specified times. Right now we are burning alot of unncessary time in operations outputting analysis restarts for hours which are not needed because we are tied to certain frequencies.

barlage:
@jmccreight I think your 1. could be tied into the =0 issue quite easily. Both would require minimal changes and could be packed into the non-positive number approach for monthly, i.e. 0=none, -1=monthly, -2=end of run (or something like that).

2. would require some additional mods, probably adding a vector to namelist.hrldas of date strings or hours (this would be easier) when restarts are requested; this would not be too involved. 

Note: I am speaking from the LSM restart viewpoint only. Given the cost savings potential, if we think this is a priority, it would not take long to do (a few hours of alone time).

PR #226, opened 2018-11-14
Metadata update

jmccreight:


jmccreight:
This should be the last update before PROD. 

Review the versioning to make sure we are on the right track and the next step: https://github.com/NCAR/wrf_hydro_nwm_public/wiki/v5.1.0-(=-NWM-v2.0)-Versioning

jmccreight:
Yes, the namelist changes are simply for testing.

PR #228, opened 2018-11-23
Testing change to work with updates to wrfhydropy

jmccreight:
THis is for testing at this point.


jmccreight:
Looks like these changes pass with the mythological wrfhydropy v0.15 🦄and with the current  wrfhydropy master. They should be merged.

jmccreight:
I should comment on why these changes are necessary. There was a change to wrfhydropy so that when a job's times are updated (start time, in particular) it preserves the path information to the restart files but updates the filenames to the new times. It used to ditch the path info and assume the new restart files were in the working directory. So this change is explicitly setting the files to be in the run directory, since that is no longer the default.

PR #244, opened 2019-01-19
Machine learning Module

jdmattern-noaa:
Mehdi Rezaeianzadeh, Nels Frazier, Donald Johnson, and I, David Mattern designed and implemented the Machine Learning module, which builds off of the same class structure and architecture of the Level Pool module. As in Level Pool, reservoirs in this module are instantiated into objects. With this module turned on, the majority of the set of total reservoirs will still be instantiated as Level Pool type reservoirs, and a subset of reservoirs determined by the Lake Parameter, Q_TYPE, will be instantiated as Machine Learning type reservoirs. These reservoirs will use previously trained machine learning models to simulate reservoir release as opposed to level pool calculations.

We have reviewed, tested, and validated the results extensively with multi-year simulations on ACF River Basin. Some other tests on Colorado Headwaters Basin confirmed that the Machine Learning Type reservoirs work on that second basin without any issues, as expected. After a successful delivery of this module for the operational NWM V2.1, we will work to have a more generic version in terms of the architecture of the machine learning model for the next year release of the NWM.

This PR has the Machine Learning option turned off, so all reservoirs would be set to Level Pool. We will then update this PR with the Machine Learning option turned on, which would need to read in three required NetCDF parameter files, one for the Machine Learning reservoir parameters, and two for the Machine Learning model parameters. We added corresponding entries for these files in the hydro.namelist. We will coordinate with James McCreight and Ryan Cabell to update the testing domain. We will soon submit a separate PR for the relevant reservoir data/states to be written to and read from their own reservoir restart files. Afterwards, we will update this PR for the Machine Learning type reservoirs to write and read relevant data to and from those restart files.


jmccreight:
I know this one is lowest priority right now, I'm going to close it and it can be revived later.

PR #203, opened 2018-10-02
Pull Request for Level Pool Module - Updated

jdmattern-noaa:
This is an updated PR from my previous last Wednesday. I created a level-pool branch where I removed my in progress machine learning modules and other miscellaneous files.

I designed and implemented the Level Pool module that creates Reservoir objects that use the Level Pool calculations within Routing. I have followed Donald Johnson and Nels Frazier's guidance for modularization with the same architecture that Donald has been using for his modules. My code development has been frequently reviewed by them. I have completed unit testing of my module and sub-modules. I also passed all of the regression tests for the nwm_ana and gridded configurations that I ran through NCAR's Docker container.

PR #251, opened 2019-02-20
Update .gitignore to ignore Run directory

rcabell:
The Run directory is a compilation artifact and should not be version controlled


jmccreight:
I was just adding this change to #203. You think it's better in a separate PR?

rcabell:
@jmccreight Yes, because it's a separate engineering issue from the Level Pool PR.

PR #254, opened 2019-02-25
Summit has

aheldmyer:


jmccreight:
This is a new configuration for using intel compilers on University of Colorado's "summit haswell" compile clusters.

PR #256, opened 2019-03-05
FORD 

jmccreight:
Create graphs and a navigable webpage for the code using ford and wrf_hydro_docker:ford

we should really merge this... open to editing a bunch of stuff, but we have enough development work that the output of ford can be very useful.


jmccreight:
@rcabell any other way you can think around having all these include files in that dir?

jmccreight:
Could we stash ford stuff at some other level? Is it ok here?

PR #257, opened 2019-03-05
Create an LSM-only configuration

### LINKED ISSUE #255: Create an LSM-only configuration
There is a need to have an LSM-only configuration of WRF-Hydro. The need has arisen because there are several outside collaborators who are funded through JTTI or OAR to work on the land model. 

## Expected Behavior

When activated, WRF-Hydro will run only the LSM. There is no dependence on Hydro inputs. This also allows single_point or vector runs for efficient development and testing.

## Current Behavior

Current configuration requires WRF_HYDRO to be active and therefore requires all the Hydro inputs.

barlage:
When activated through setting the environment variable LSM_ONLY, WRF-Hydro will run only the LSM. There is no dependence on Hydro inputs. This also allows single_point or vector runs for efficient development and testing.

Fixes #255

### DIFF
--- a/trunk/NDHMS/Land_models/NoahMP/IO_code/main_hrldas_driver.F
+++ b/trunk/NDHMS/Land_models/NoahMP/IO_code/main_hrldas_driver.F
@@ -42,10 +42,8 @@ program Noah_hrldas_driver
 #endif
        call land_driver_exe(ITIME)
    end do
-#ifdef WRF_HYDRO
-   call hydro_finish() 
-#endif
 
+   call hydro_finish() 
 
 END 
 
--- a/trunk/NDHMS/Land_models/NoahMP/IO_code/module_NoahMP_hrldas_driver.F
+++ b/trunk/NDHMS/Land_models/NoahMP/IO_code/module_NoahMP_hrldas_driver.F
@@ -553,6 +553,7 @@ module module_NoahMP_hrldas_driver
   IOPT_RSF = surface_resistance_option
 
 
+#ifdef WRF_HYDRO
   !!----------------------------------------------------------------------------
   !! channel_only
   call updateNameList("channel_only",0)
@@ -599,6 +600,7 @@ module module_NoahMP_hrldas_driver
   endif
   !!----------------------------------------------------------------------------
   !! channel_only
+#endif
 
 
 !---------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


jmccreight:
We will have to come up for some tests for this, should we accept it!

barlage:
@jmccreight I'm sure there is some document I should be reading, but are you referring to automated testing? If yes, then slightly modified existing tests should work. This configuration should use the same forcing and wrfinput files, the difference being that only the LSMOUT files need to be checked.

jmccreight:
Yes, I'm referring to automated testing. The tests are self-documenting! ;) I can help with this. I'll try to get to this soon

barlage:
OK, I've let some collaborators know that this branch exists on my fork so we may get some external and internal practical testing done as well.

jmccreight:
We should check that the code compiles with that env var set and only produces LDASOUTS through some desired date. I guess we can skip perfect restart and ncores test as those are exercised with the full model.
Anything else?

barlage:
If we want this to officially support single-point and vector, we should have some tests on that. I'm checking some ncores on my vector forcing now. Perfect restart should still work regardless of the domain structure.

barlage:
In what I can only assume is an act of retaliation, @aubreyd has suggested that LSM_ONLY is really a patch for the continued life of the WRF_HYDRO env variable. Many years ago WRF_HYDRO was added since WRF_Hydro was added into an existing LSM_ONLY HRLDAS. 

Since this environment is now WRF-Hydro only, is the better solution to finally remove WRF_HYDRO? And then wrap the currently necessary WRF_HYDROs in not-LSM_ONLYs. Is there a downside to this in coupled mode? @kafitzgerald

rcabell:
@barlage I think this is reasonable, as the `#ifdefs` really should be wrapping the exceptional cases, not the common use cases. That being said, I'd wait to do that until after the 2.1 delivery just to be on the safe side. I'll let other chime in with their opinions on that or other timelines for making this change.

aubreyd:
Definitely retaliation.

For clarification, my thought was not to remove the WRF_HYDRO ifdefs from the code. Instead, remove it from the required environment variables and replace it with the new LSM_ONLY one. Then the compile script would set WRF_HYDRO depending on the LSM_ONLY setting (I think what Mike is already doing). Current compile script requires WRF_HYDRO=1, which makes the env var sort of useless unless there is some hidden use (other than LSM only) that I am unaware of. This is mostly semantics as opposed to real code changes.

kafitzgerald:
I think either of these approaches would work.  WRF_HYDRO will still be an environment variable for WRF, but if we don't need it or want to set an appropriate default on the hydro side I think that's fine.  Definitely would like to test though so keep me in the loop.

aubreyd:
If still needed for WRF, we can certainly keep it exposed. Just looking for an opportunity to avoid the training question when we review setEnvar and the students inevitably ask: Why is there an environment variable where nothing works if it is not set to 1??

kafitzgerald:
I think it can be removed and just mentioned more explicitly in the coupled build instructions (that way we put the info where/when folks need it).  I agreed that it's probably confusing to mix it with the WRF-Hydro compile time options.

jmccreight:
Per discussion with @barlage we are putting this on ice until after 2.1. It will be available as a branch on his fork. Later we will unify the everything

PR #260, opened 2019-03-12
New wrfhydropy

jmccreight:
This will require a pypi release. 
This picks up changes to testing to adapt to / benefit from changes in wrf_hydro_py.


jmccreight:
This needs a pypi release to pass, and a rebulid of the docker container with that release (I dont think it autoupdates the wrfhydropy version when it runs, but it might).

jmccreight:
done via v5.1.0 merges to master and the PRs #267 and #270 
closing

PR #262, opened 2019-03-15
Channel-route output metadata fixes when scale_factor and add_offset …

jmccreight:
…are applied.

This should wait until testing can be applied to check NWM outuputs and their metadata.

PR #263, opened 2019-03-18
Chrt metadata

logankarsten:
… for fill/valid ranges on other output routines
1.) Reverting feature_id long name on CHRTOUT files to v1.2 per OWP request. 
2.) Applying previous PR for CHRTOUT valid min/max/fill to all output routines to ensure correct integer valid min/max and fill values are going into output files. 
3.) Scaling scale_factors for variables appropriately to ensure 4-byte integer limits are not exceeded. 
4.) Fixing units on velocity in CHRTOUT.
5.) Removing unused vis_nir and soil_layers_stag dimensions in LDASOUT and RTOUT files when related variables are not produced.


jmccreight:
to merge before #262. But after the metadata tests are merged (in a forthcoming PR).

jmccreight:
The results show the expected failures. We will review this again when it merges with master

PR #266, opened 2019-03-19
v5.1.0: Testing improvements to cover NWM output metadata

jmccreight:


jmccreight:
There errors are expected as the reference does not have the new configurations. Merging

PR #275, opened 2019-03-26
feature: groundwater bucket loss option

kafitzgerald:
This PR adds a namelist option (bucket_loss) to allow for losses from the groundwater bucket defined as a fraction of bucket outflow.  By default, this option is turned off and code should pass regression tests.  However, in order to run this configuration an additional variable (Loss) is required in the GWBUCKPARM.nc parameter file.  I'm happy to generate that file for the test domain(s) and create a new test for this configuration, but figured I would open this PR first to start that discussion.


kafitzgerald:
Looks like managed to add a typo when fixing merge conflicts, but that should be fixed now.  

It will still fail on comparing the *GWOUT* files though.  I had forgotten that I changed the variable name from "depth" -> "level".  I can remove this change for now if that makes more sense as a separate PR.  However, I think we should make the change (and hopefully sooner rather than later).  The current variable name is pretty misleading and given that these are not output for NWM (currently) the change shouldn't be too disruptive.

trunk/NDHMS/Routing/module_GW_baseflow.F
```diff
@@ -90,6 +95,9 @@ subroutine simp_gw_buck_nhd(           &
 !!!Initialize variables...
    net_perc = 0.
    qout_gwsubbas = 0.
+   if (bucket_loss .eq. 1) then
+      qloss_gwsubbas = 0.
```
jmccreight: We really need a constants module...

rcabell: @jmccreight agreed, but 0 and 1 are the _least_ magic numbers...

trunk/NDHMS/Routing/module_NWM_io.F
```diff
@@ -5026,11 +5032,17 @@ subroutine output_gw_NWM(domainId,iGrid)
...
+             fileMeta%outFlag(3) = 1
+             call gw_write_io_real(rt_domain(domainId)%numbasns,rt_domain(domainId)%qloss_gwsubbas, &
+                                   rt_domain(domainId)%basnsInd,g_qloss_gwsubbas)
+         endif
       endif
 
    else
       allocate(g_qin_gwsubbas(rt_domain(domainId)%gnumbasns))
       allocate(g_qout_gwsubbas(rt_domain(domainId)%gnumbasns))
+      allocate(g_qloss_gwsubbas(rt_domain(domainId)%gnumbasns))
```
jmccreight: relying on this going out of scope? I would expect to see a new, accompanying deallocate...

kafitzgerald: Good catch.  Will fix this.

trunk/NDHMS/Routing/module_NWM_io_dict.F
```diff
@@ -2496,18 +2496,18 @@ subroutine initGwDict(gwOutDict)
...
-   gwOutDict%timeZeroFlag(:) = [0,0,1]
-   gwOutDict%fillReal(:) = [-9999.0,-9999.0,-9999.0]
-   gwOutDict%missingReal(:) = [-9999.0,-9999.0,-9999.0]
-   gwOutDict%validMinDbl(:) = [-10000.0d0, -10000.0d0, -10000.0d0]
-   gwOutDict%validMaxDbl(:) = [10000.0d0, 10000.0d0, 10000.0d0]
+   gwOutDict%varNames(:) = [character(len=64) :: 'inflow','outflow','loss','depth']
+   gwOutDict%longName(:) = [character(len=64) :: 'Bucket Inflow','Bucket Outflow','Bucket Loss','Bucket Depth']
+   gwOutDict%units(:) = [character(len=64) :: 'm3 s-1','m3 s-1','m3 s-1','mm']
+   !gwOutDict%coordNames(:) = [character(len=64) :: 'latitude longitude','latitude longitude','latitude longitude','latitude longitude']
+   gwOutDict%scaleFactor(:) = [0.00001,0.00001,0.00001,0.1]
```
jmccreight: Is this high, or necessary to avoid step functions?

trunk/NDHMS/Routing/module_RT.F
```diff
@@ -1165,6 +1165,14 @@ subroutine LandRT_ini(did)
       allocate (rt_domain(did)%z_max (rt_domain(did)%numbasns))
       allocate (rt_domain(did)%basns_area (rt_domain(did)%numbasns))
 
+      if(nlst_rt(did)%bucket_loss .eq. 1) then
+         allocate(rt_domain(did)%qloss_gwsubbas(rt_domain(did)%numbasns))
```
jmccreight: do these deallocate? I prefer explicit deallocation. it could be very useful at some later point.

trunk/NDHMS/Routing/module_GW_baseflow.F
jmccreight: mere mortal numbers, especially if they are integers...

jmccreight:
Is this correct: 
the output now includes the instantaneous amount lost but no accumulated amount lost
?
That's fine. I just wanted to clarify.

trunk/NDHMS/Routing/module_RT.F
kafitzgerald: They don't, but nor do the other similar arrays.

kafitzgerald:
It is correct currently, but given the rounding errors on the outputs I think I need to add the accumulated outputs as well.

trunk/NDHMS/Routing/module_NWM_io_dict.F
kafitzgerald: I think something between what it was and these (picked because they are consistent w/ chrt vars) will work.  Thought I had taken these out with the metadata changes and was planning to adjust anyway.

jmccreight:
I'm not advocating for an accumulated field. I think it's not trivial to implement one. It might be easier because the accumulation is only an output diagnostic... but we are also not accumulating the inflows and outflows of the bucket. I just wanted to clarify what was happening.

rcabell:
@kafitzgerald, @jmccreight  -- what's the current status of this? It hasn't been touched in 3 weeks. Is it ready to go?

kafitzgerald:
> It is correct currently, but given the rounding errors on the outputs I think I need to add the accumulated outputs as well.

Per our conversation offline.  I'll wait on this.

kafitzgerald:
> @kafitzgerald, @jmccreight -- what's the current status of this? It hasn't been touched in 3 weeks. Is it ready to go?

I'll make that change to scale factor right now (re: comment from James) and then I think we're good to go for this PR.  

We'll need to update testing with new files for the loss option after the merge.  I'll chat w/ you and @jmccreight offline about how we want to do that.

Issue #232, opened 2018-11-29
duplicating MPI communicators


jmccreight:
https://github.com/NCAR/wrf_hydro_nwm_public/blob/8e1dbc21aca54dcccb6cac68d7c009d3e3a1926d/trunk/NDHMS/MPP/mpp_land.F#L156

The above is how not to do it, should use MPI_COMM_DUP.

glennpj:
Hi,
I am building the coupled model of WRF/WRF-hydro for one of our researchers. I am seeing the following error when trying to run the front_range_CO test case, which perhaps is related to the comment above.
```
*** An error occurred in MPI_Bcast
*** reported by process [47923237289985,1]
*** on communicator MPI_COMM_WORLD
*** MPI_ERR_COMM: invalid communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```
I did a quick search in MPI docs and see,

> In particular, no library routine should use MPI_COMM_WORLD as the communicator; instead, a duplicate of a user-specified communicator should always be used. 

https://www.mpich.org/static/docs/latest/www3/MPI_Comm_dup.html

I may be looking at the wrong thing but this was the most promising lead that I found. How would I go about duplicating the communicator?

Thanks.

afanfa:
a quick fix is the following:

` integer :: ierr
call MPI_COMM_DUP(MPI_COMM_WORLD, HYDRO_COMM_WORLD, ierr)
`

assuming that an integer varia

glennpj:
@afanfa Sorry for the delayed response but that seems to have done the trick. I made the change in mpp_land.F and CPL_WRF.F.

Thanks so much.

afanfa:
@glennpj that's great news. @jmccreight I am going to make a pull request to reflect that change on master as well.

jmccreight:
#281

PR #279, opened 2019-04-05
Nudging slowdown/deadlock fix

jmccreight:
…er true 2-sided mpi.

2. Perform proper MPI communicator duplication.
3. Clean up some errors that affect scheduled runs not being waited for.

### DIFF
--- a/tests/test_1_fundamental.py
+++ b/tests/test_1_fundamental.py
@@ -101,13 +101,13 @@ def test_run_candidate(candidate_sim, output_dir, ncores, exe_cmd):
     # job_id='test_job'
     wait_job(candidate_sim_copy)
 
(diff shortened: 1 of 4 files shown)


jmccreight:
Addresses/closes #232

jmccreight:
The nudging code was causing slowdowns under real 2-sided MPI (early MPT and OpenMPI are doing 1-sided under the hood, so were not revealing this problem). 
I assume this is what caused the deadlock observed on WCOSS. 
Root cause was simply a `continue` statement where there should have been a `cycle` statement in a loop executing MPI. Resulting time in loop was reduced by orders of magnitude, the nudgingLastObs files are now generally done writing withing 3s of the HYDRO_RST file completion.

PR #281, opened 2019-04-05
V5.1.0: merge conflicts with MPI comms with mastser and cleanup

jmccreight:


jmccreight:
This follows #279 where merging it to master resulted in merge conflicts. I consulted with @afanfa @danrosen25 about the best way to do that. Also resolves #232 when merged to master.

PR #286, opened 2019-04-05
Adding a first cut cheyenne-specific bash script for calling pytest. …

jmccreight:
…Go DI!

### DIFF
--- a/tests/local/cheyenne/model_test.sh
+++ b/tests/local/cheyenne/model_test.sh
@@ -1,0 +1,197 @@
+#!/bin/bash
+# A first productiony test script for cheyenne.
+# James McCreight
+# April 5, 2019
+# Why bash?
+#    1) might need to establish the python env from scratch,
+#    2) set the modules is much easier
+#    3) git is easier in bash.
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+the_help="\
+Model Testing
+
+Required
+    -c: 
+        Candidate directory path. The code to test. 
+    -r:
+        Reference directory path. The code to test against.
+
+Options:
+    --reference_update: 
+       default=true
+       In the reference repo, fetch upstream (NCAR) and checkout branch matching 
+       that of the candidate?
+    --compiler: 
+        [default=ifort, gfort, <You can try whatever is available in modules>] 
+        Currently:  ifort -> intel/17.0.1   and  gnu -> gnu/7.1.0
+        We should try to keep this update along with what is here:
+        https://github.com/NCAR/wrf_hydro_nwm_public/wiki/Compiler-requirements-by-version
+    --mpi: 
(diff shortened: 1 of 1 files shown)


jmccreight:
This script is super general purpose. 
There are a few rough edges
1) option to use standardized python env
2) gymnastics around the reference repo... not ideal currently (but convenience)
3) not currently wired the exe_cmd through for non-standard MPI (easy).

tests/local/cheyenne/model_test.sh
```diff
@@ -0,0 +1,197 @@
...
+## Candidates branch to tag the test directory and optionally update the reference.
+branch_name="$(git symbolic-ref HEAD 2>/dev/null)" || branch_name="$(git rev-parse --short HEAD)"
+branch_name=${branch_name##refs/heads/}
+printf "\e[7;49;94mTesting branch: $branch_name in $candidate_dir\e[0m\n"
+output_dir=/glade/scratch/`whoami`/take_test_$(basename $domain_dir)_${branch_name}
+
+#-------------------------------------------------------
+# "Update" the reference dir? This makes the branch match that of
+# the candidate and it fetches the NCAR:branch_name
+# There are two potential pitfalls here,
```
kafitzgerald: I'm a little bit confused about why the default behavior here was chosen.  The first pitfall seems like it'll be pretty common and the automatic updating makes some assumptions for how repos are configured.

jmccreight: It's just what was convenient for me... it was good for most of my cases. Should be good for most CI-ish cases. But I'm open to better ideas

rcabell: I think ultimately this should be an overridable default. But ok as a starting point.

tests/local/cheyenne/model_test.sh
```diff
@@ -0,0 +1,197 @@
...
+printf "\e[7;49;94mModule information\e[0m\n"
+module purge
+# Is this strict enough in the sense that things might be changing?
+module load  $compiler_module  $mpi  ncarcompilers  netcdf  ncarenv || exit 4
+module list
+
+#-------------------------------------------------------
+# Python Env
+# Need option to install standardized python env.
+# Look in to pipenv some day and a required.txt file or similar
```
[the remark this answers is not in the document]
rcabell: I think it's not a huge performance hit to create a venv and tear it down each time, with the benefit of reproducibility being worth the download/install time.

tests/local/cheyenne/model_test.sh
jmccreight: Agreed.

tests/local/cheyenne/model_test.sh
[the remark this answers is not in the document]
jmccreight: We are going to see if I can use katelyns env, then we can just point to that on cheyenne and maintain that. A bit speedier...

PR #289, opened 2019-04-09
Small but big improvements to cheyenne test script

jmccreight:


kafitzgerald:
Ah, shoot.  Thought that was just going to be a template / test case.  Here's a replacement in a safer location: /glade/u/home/katelynw/python/envs/testing

Issue #290, opened 2019-04-24
Level Pool Mass Balance Bug


hellkite500:
The level pool formulation doesn't correctly conserve mass when exceeding max conditionions.

## Expected Behavior
![correct](https://user-images.githubusercontent.com/5186647/56674950-9d27f780-6680-11e9-947e-5d87d28a99c5.png)

When a reservoir is below max, and the current conditions would cause it to exceed max, it should fill the available storage and release the excess water.

## Current Behavior
![broke](https://user-images.githubusercontent.com/5186647/56674974-a7e28c80-6680-11e9-8271-55e5ccc671bd.png)

When current conditions cause the reservoir to exceed max, storage becomes max and all inflow is passed to outflow.  This creates water in the reservoir!

## Possible Solution
#259 contains the one line fix for this issue. (Commit 674a4750c4dc56df7169336fa2eefac11e594aba)
When the reservoir exceeds max, the discharge should be
`discharge = qi1 - (((maxh - Htmp) * sap) / dt)`

When the reservoir was not full, but would exceed, `maxh-Htmp = available storage` and only the excess inflow goes to outflow.

When the reservoir is already full, `maxh-htmp = 0`, and all inflow goes to outflow.

hellkite500:
As has been realized, this fix DOES NOT address what happens to orifice + weir outflows in the overtop condition.  The proposed solution doesn't account for these, but could be modified to do so.  

```discharge = qi1 - (((maxh - Htmp) * sap) / dt) + (orifice + weir)
H = <recompute based on Htmp, discharge, qi1>```
Also need to appropriately recompute H (it may not longer be maxh since the extra volume is removed from the reservoir.)  This would likely mean that we maintain NEARLY max elevation, minus the extent of the weir to move flow, so by the next time step we are below max again, then will cross over with the new inflow, move the excess and however much the weir can handle, and the elevation is once again below max.

jmccreight:
The file attached to #292 adds this further detail:

## Modification on the Level Pool Method
This is an example of how the new fix would impact the outflow simulated by the Level Pool (LP) method. We added here results for the lake_id ==16944276. Note that graphs may not clearly represent the difference.
[...]

PR #292, opened 2019-05-01
Level Pool Bug Fix

jdmattern-noaa:
This is the Level Pool bug fix for instances when the calculated water elevation exceeds the max elevation. Attached are the results and explanation of a unit test for this case. This unit test is also in the PR. To reproduce the results, go to the reservoir directory in the terminal, type 'make test', and then './reservoir_tests'. 
Addresses #290 (see that issue for more details).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ trunk/NDHMS/HRLDAS/user_build_options
 /.project
 trunk/NDHMS/Run
 trunk/NDHMS/Run/
-trunk/NDHMS/setEnvar.sh
 trunk/NDHMS/Makefile.comm
 trunk/NDHMS/Land_models/NoahMP/user_build_options
 trunk/NDHMS/Land_models/Noah/user_build_options
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ trunk/NDHMS/HRLDAS/user_build_options
 /.project
 trunk/NDHMS/Run
 trunk/NDHMS/Run/
-trunk/NDHMS/NWM/
 trunk/NDHMS/setEnvar.sh
 trunk/NDHMS/Makefile.comm
 trunk/NDHMS/Land_models/NoahMP/user_build_options
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ trunk/NDHMS/HRLDAS/user_build_options
 /.project
 trunk/NDHMS/Run
 trunk/NDHMS/Run/
+trunk/NDHMS/NWM/
+trunk/NDHMS/setEnvar.sh
 trunk/NDHMS/Makefile.comm
 trunk/NDHMS/Land_models/NoahMP/user_build_options
 trunk/NDHMS/Land_models/Noah/user_build_options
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -12,6 +12,8 @@ trunk/NDHMS/HRLDAS/user_build_options
 /.project
 trunk/NDHMS/Run
 trunk/NDHMS/Run/
+trunk/NDHMS/NWM/
```
jmccreight: This looks custom. Unless I'm mistaken. It could be removed and placed in an untracked .gitignore inside the/your untracked trunk/NDHMS/NWM repository

laurareads:
I'm working on a CONUS science test to confirm the results.

trunk/NDHMS/Routing/Reservoirs/reservoir_tests.F
```diff
@@ -5,9 +5,24 @@ program reservoir_unit_tests
     implicit none
```
jmccreight: How is this built?

.gitignore
jdmattern-noaa: Okay thanks. I will change this.

trunk/NDHMS/Routing/Reservoirs/reservoir_tests.F
jmccreight: I see it in your comments on this PR on how to make and run. Can you please add those instructions to the top of this file?

It looks like this file was there previously, so this should probably have been done previously....

jmccreight:
Should this PR reference #290 ? Does it address #290, does it close #290?

trunk/NDHMS/Routing/Reservoirs/reservoir_tests.F
jdmattern-noaa: Type 'make test' in the Reservoir directory. Then run with './reservoir_tests'.

.gitignore
jmccreight: Not a biggie, but every line matters and i'm trying to build peoples habits of carefully reviewing their git diffs for PRs. Doing this, building this practice should help result in "bite sized" PRs.

trunk/NDHMS/Routing/Reservoirs/reservoir_tests.F
jmccreight: Functionally, this PR is not that closely tied to the code, it's at least 1 step away. Putting the instructions at the top of the file will help close that gap.

jdmattern-noaa: Okay will do.

jmccreight:
Also, can the contents of the attached file be added to #290? It will provide more information about the issue in 1 place. I'm glad to do it if @jdmattern-noaa confirms this is appropriate.

jdmattern-noaa:
Yes this fixes #290. Yes if you don't mind attaching it.

.gitignore
jmccreight: I would also say that historically, and I would say continuing, we have a bad habbit of letting in too much unnecessary code. This is probably even more about the reviewers than the PR.

jdmattern-noaa:
And closes #290

.gitignore
jdmattern-noaa: Good points. I will do that from now on.

hellkite500:
Also addresses #312 by 79111adedd06a058ff8150b62d57f4e3bbd6b4db

rcabell:
_**The code is not passing internal Science Tests**_ 

@laurareads will follow up with more details...

jmccreight:
This code/approach was abandoned in favor of a different appraoch. 
This PR will not be reopened.

PR #297, opened 2019-05-21
Use MPP_LAND_INIT() in Noah coupling vs direct MPI init

rcabell:
**TYPE**: bug fix

**KEYWORDS**: Noah, MPI, MPP_Land

**SOURCE**: Ryan Cabell (NCAR)

**DESCRIPTION OF CHANGES:**

The Noah LSM coupling used a direct, private initialization of MPI that caused conflict with the new HYDRO_COMM_WORLD facility that was added recently to `mpp_land.F`. Not using the `MPP_LAND_INIT()` leaves the new private communicator uninitialized, leading to a segfault when attempting to access it later.

This commit updates `Noah_hrldas_driver.F` to use the new common MPP/MPI initialization scheme, `mpp_land.F` to include error handling that was being done in the Noah driver, and an update to the `Noah/IO_Code Makefile` to include the MPP module.

Produces no answer changes.

**LIST OF MODIFIED FILES:**

trunk/NDHMS/Land_models/Noah/IO_code/Makefile
trunk/NDHMS/Land_models/Noah/IO_code/Noah_hrldas_driver.F
trunk/NDHMS/MPP/mpp_land.F

**TESTS CONDUCTED:**

Noah runs complete without crashing when using the uninitialized `HYDRO_COMM_WORLD`

### DIFF
--- a/trunk/NDHMS/Land_models/Noah/IO_code/Makefile
+++ b/trunk/NDHMS/Land_models/Noah/IO_code/Makefile
@@ -18,7 +18,7 @@ Noah_hrldas_driver.o: Noah_hrldas_driver.F
 	$(RM) $(*).f
 	$(CPP) $(CPPFLAGS) $(CPPHRLDAS) $(*).F > $(*).f
 	$(COMPILERF90) -o $(@) -c $(F90FLAGS) $(FREESOURCE) $(MODFLAG). \
-	$(MODFLAG)../Noah $(MODFLAG)../Utility_routines $(NETCDFMOD) $(*).f
+	$(MODFLAG)../Noah $(MODFLAG)../Utility_routines -I../../../MPP $(NETCDFMOD) $(*).f
 	$(RM) $(*).f
 	@echo ""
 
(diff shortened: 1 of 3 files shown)


afanfa:
Looks good to me

jmccreight:
I'm very confused about how things got into this state and how this bug was detected or not detected. How did it pass CI? (Was there a problem with CI?) Were there specific conditions which reveal this bug?

It looks like 
```
call MPI_COMM_DUP(MPI_COMM_WORLD, HYDRO_COMM_WORLD, ierr)
```
is what caused the issue? The duplicated communicator was the one to use and not MPI_COMM_WORLD. Then hydro and LSM had separate/different communicators which should have been the same?

It's also remarkable that code reuse `MPP_LAND_INIT()` solves the problem! Always a sign of improvement is elimination of code.

rcabell:
James, it only shows up in Noah because Noah was doing an MPI init on its own, so when MPP_LAND did its check it saw it was already init'ed and the MPI_COMM_DUP wasn't being called, and the value of that variable was undefined. It got used later and boom, segfault.

Correct me if I'm wrong, but the CI only tests NoahMP, right?


________________________________

jmccreight:
Ah yes. Sorry, I didnt notice that this was Noah. I'm biased too! OK. OK, this makes sense now.

jmccreight:
And yes, CI only currently tests NoahMP. IT would be fairly easy to add Noah however.

PR #300, opened 2019-05-23
fix the cheyenne python env path to something not on scratch.

jmccreight:
@kafitzgerald


jmccreight:
I have one more small change to the script to add. ...

PR #306, opened 2019-06-03
Fixes to the overland routing off configuration.

aubreyd:
Answer changes only when overland routing is off (affects UDMP and non-UDMP). Moved surface runoff that was mapped through GW buckets in non-UDMP configs to behave as pass-through only (as opposed to bucket attenuation; ugdrunoff is still through bucket). Both are combined for gwout. Confirmed ncores and perfect restarts. Needs at least one new output var to fully track all WB components (TBD).

### DIFF
--- a/trunk/NDHMS/Routing/module_GW_baseflow.F
+++ b/trunk/NDHMS/Routing/module_GW_baseflow.F
@@ -140,10 +140,6 @@ contains
                    ii =  LUDRSL(k)%cell_i(m)
                    jj =  LUDRSL(k)%cell_j(m)
                    if(ii .gt. 0 .and. jj .gt. 0) then
-                      if(OVRTSWCRT.ne.1) then
-                           LQLateral(k) = LQLateral(k)+runoff1x(ii,jj)*LUDRSL(k)%cellWeight(m)/1000 &
-                               *cellArea(ii,jj)
-                      endif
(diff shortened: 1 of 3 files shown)


aubreyd:
CI testing results are as expected - answer changes in hydro outputs and restart in NWM long-range configuration only (only tested configuration where OVRTSWCRT=0).

trunk/NDHMS/Routing/module_GW_baseflow.F
```diff
@@ -339,9 +339,11 @@ subroutine simp_gw_buck(ix,jx,ixrt,jxrt,numbasns,gnumbasns,basns_area,basnsInd,g
 
 #ifdef MPP_LAND
     call gw_sum_real(sum_perc8,numbasns,gnumbasns,basnsInd)
+    call gw_sum_real(sum_perc8_surf,numbasns,gnumbasns,basnsInd)
     call gw_sum_real(ct_bas8,numbasns,gnumbasns,basnsInd)
 #endif
    sum_perc = sum_perc8
+   sum_perc_surf = sum_perc8_surf
```
jmccreight: why do it in double precision and then throw it away?

jmccreight: there's probably `sum`thing hiding in `gw_sum_real`

aubreyd: No idea - I just duplicated what was already in the code. Pre-dates me.

PR #320, opened 2019-06-20
Noah-MP pedotransfer function addition

barlage:
TYPE: enhancement, eventually

KEYWORDS: Noah-MP, soil model

SOURCE: Michael Barlage (NCAR)

DESCRIPTION OF CHANGES:

Add the Noah-MP pedotransfer function subroutine to the Noah-MP module with minimal changes to higher-level code. The pedotransfer code is not active; this is a commit to bring the Noah-MP code up to the community code.

Produces no answer changes.

This PR addresses 4. in #333 

LIST OF MODIFIED FILES:
M module_sf_noahmplsm.F
M module_sf_noahmpdrv.F
M MPTABLE.TBL
M module_NoahMP_hrldas_driver.F

TESTS CONDUCTED:

1. One-year simulation (Cheyenne, PGI) in two calibration basins produces no changes


trunk/NDHMS/Land_models/NoahMP/phys/module_sf_noahmpdrv.F
```diff
@@ -1212,6 +1214,122 @@ SUBROUTINE TRANSFER_MP_PARAMETERS(VEGTYPE,SOILTYPE,SLOPETYPE,SOILCOLOR,CROPTYPE,
 
  END SUBROUTINE TRANSFER_MP_PARAMETERS
 
+SUBROUTINE PEDOTRANSFER_SR2006(nsoil,sand,clay,orgm,parameters)
```
jmccreight: SR2006? Maybe drop in what that is. A comment with the citation.

jmccreight: Saxton and Rawls (2006)

trunk/NDHMS/Land_models/NoahMP/run/MPTABLE.TBL
```diff
@@ -471,3 +471,60 @@ GRAINPT_S8 =    0.0,    0.0,    0.0,    0.0,    0.0,
 BIO2LAI   =  0.035,  0.015,  0.015,  0.015,  0.015,  ! leaf are per living leaf biomass [m^2/kg]
 
 /
+
+&noahmp_optional_parameters
```
jmccreight: Really sticking with the text tables.

jmccreight:
I think we should open an issue about this feature and add this PR to its evolution. 
I know that it is more NoahMP than WRF-Hydro... would be great if that issue were just in the NoahMP repo.

trunk/NDHMS/Land_models/NoahMP/phys/module_sf_noahmpdrv.F
barlage: @jmccreight I do have the reference in the option definition. Since this is an attempt to unify with external codes, I would prefer to hold off on an addition like this until after the release and/or when we make this part of the code active.

trunk/NDHMS/Land_models/NoahMP/run/MPTABLE.TBL
barlage: NetCDF someday, maybe. Namelists are a nice way to avoid dependencies...

rcabell: Is there a reason to use unit=15? My only concern here is that 15 is used by WCOSS as the standard output unit. If we know for sure that this will never be used there, not a problem, but otherwise could be an issue.

barlage: @rcabell No reason to use unit=15. Just looked back through old code and it's been that since the first version I have from 2011.

PR #321, opened 2019-06-24
Added nwm_doxygen file that builds nwm documentation, included call a…

donaldwj:
…nd called by graphs; Added nwm.md and overland.md mark down pages, these pages where for internal reference in refactoring and need to be updated for use with the complete model.

The doxygen file provided builds documentation and call/called by graphs for all function and files in the current NWM source.  However the top level markdown file that creates the index.html needs to be updated and additional markdown pages may need to be added to make the resulting documentation easier to use.

Donald Johnson


donaldwj:
The parts I documented in more depth where NWM specific one of the reasons
to merge this is so that the rest of the model can be similarly documented
by those with the requisite knowledge. Eventually it would be good for any
directory in the source tree to have markdown file that gives the high
level documentation for that part of the code. A different top level mark
down for the inital landing page is also needed as the existing one was
designed for internal use of the refactoring project.

kafitzgerald:
Ok, if we want to just merge this now so it's there and then work on something a bit more general (I'm happy to start this) that sounds good to me.  

I am curious if you want to maintain the content of the more NWM centric internal landing page going forward or if that's more of a placeholder.  FWIW I don't think it's bad idea to have that type of content in the repository.  Just not sure where it sure land.

donaldwj:
The current landing page is a placeholder and eventual top level landing
page should probably include a NWM link among other links

kafitzgerald:
Ok, great.  That's more or less what I was thinking.

PR #322, opened 2019-06-25
Nwm cmake

donaldwj:
This is the initial attempt to build WRF_Hydro/NWM with cmake. It works for my local tests and i tried to include capabilities present in the existing builds system.

To use do the following 
create a build directory in trunk/NDHMS. 'mkdir cmake-build'
change curent to directory to the build directory. 'cd cmake-build'
'cmake ..'
'make'

At the moment the cmake build should probably be maintained as a secondary build system until it can be better tested.

Donald Johnson


rcabell:
@donaldwj - I've not been able to get this working either on my local system or Cheyenne; it fails trying to find NetCDF. I've set the NETCDF_DIR environment variable as it appears necessary from `FindNetCDF.cmake`.  Are there any other env vars I'm missing here?

donaldwj:
I didn't have to set NETCDF_DIR when i tested it locally.

you can try setting

NETCDF_INCLUDES
NETCDF_LIBRARIES

directly which will short circuit the detection script


Example output from my system below
========================================================================
[dwj@localhost cmake-build]$ rm CMakeCache.txt
[dwj@localhost cmake-build]$ cmake ..
-- Found MPI_C: /usr/lib64/openmpi/lib/libmpi.so
-- Found MPI_CXX:
/usr/lib64/openmpi/lib/libmpi_cxx.so;/usr/lib64/openmpi/lib/libmpi.so
-- Found MPI_Fortran:
/usr/lib64/openmpi/lib/libmpi_usempi.so;/usr/lib64/openmpi/lib/libmpi_mpifh.so;/usr/lib64/openmpi/lib/libmpi.so

-- Found NetCDF:
/usr/lib64/libnetcdff.so;/usr/lib64/libnetcdff.so;/usr/lib64/libnetcdf.so
-- NetCDF Include Dir: /usr/include
-- Setting LSM to: NoahMP
=============================================================
-- Start of WRF-Hydro Env VARIABLES
WRF_HYDRO = 1
HYDRO_D = 0
HYDRO_REALTIME = 0
WRF_HYDRO_RAPID = 0
SPATIAL_SOIL = 0
WRFIO_NCD_LARGE_FILE_SUPPORT = 0
NCEP_WCOSS = 0
WRF_HYDRO_NUDGING = 0
OUTPUT_CHAN_CONN = 0
PRECIP_DOUBLE = 0
=============================================================
-- Using gfortran
-- Building NoahMP LSM
-- Configuring done
-- Generating done
-- Build files have been written to:
/home/dwj/projects/nwm/wrf_hydro_nwm_public/trunk/NDHMS/cmake-build

rcabell:
I'm guessing since you have NetCDF in `/usr/include` / `/usr/lib` it's one of the places CMake looks by default. For those using Modules, it needs a more explicit pointer, which isn't working. I'll try to get that `FindNetCDF.cmake` script working and commit the changes back to this PR when it does.

donaldwj:
Probably this line in FindNetCDF.cmake is the issue

find_path (NETCDF_INCLUDES netcdf.h HINTS NETCDF_DIR ENV NETCDF_DIR)

Isn't finding what is looking for in either the default directories, the
cmake variavle NETCDF_DIR or the environment variable NETCDF_DIR.

Donald Johnson

rcabell:
I've set the `NETCDF_DIR` environment variable and it's still not happy; time for me to start digging through the CMake docs, I guess!

donaldwj:
It looks like the first thing searched for is netcdf.h, find_path needs to
be able to find that file.

Donald Johnson

donaldwj:
What is it set to and where is netcdf.h relative to that path?

Sadly the FindNetCDF.cmake is found code as NetCDF detection is not built
in.

Donald Johnson

rcabell:
I got it to work finally with this:

```
find_path (NETCDF_INCLUDES netcdf.h
           PATHS "$ENV{NETCDF}/include" ENV NETCDF_INC)

find_library(NETCDF_LIBRARIES_C       NAMES netcdf
             PATHS "$ENV{NETCDF}/lib" ENV NETCDF_LIB)
```

Still some issues in other areas (missing CMakeLists.txt in `trunk/NDHMS/Routing/Reservoirs`) but making progress!

donaldwj:
I think i wrote a file for reservoirs maybe it wasnt added to the commit,
I'll check.

Donald Johnson

donaldwj:
The reservoirs CMakeLists.txt was missing from the branch I have pushed it
the branch, does that update the pull request?

donaldwj:
That is the behavior everyone is seeing it hangs if you are using the intel
compilers.

donaldwj:
Can probably be done with some use of add_custom_command to add post build
commands to the two rules (NoahMP or Noah) for building wrfhydro.exe? Would
you want the run directory to be in the cmake-build directory or in the
main source directory (thus potentially clobering files from a normal
build)?

rcabell:
@donaldwj -- it would be nice if it went to the out-of-tree build directory (if run like that), or the source root (like it traditionally did). Is it possible to switch the behaviour like that?

donaldwj:
The easiest way I see to add this would be to place the 'Run' directory
relative the the CMAKE_BUILD_DIR which depending on how cmake was called
would either be in tree (call of 'cmake' in the source directory) or out of
tree (call of 'cmake <source_root>' from another directory).

donaldwj:
Most recent push should correct that.

rcabell:
I pushed a few changes to this which staled out @kafitzgerald's review.  You can see the updates in 1c518c4 if you so desire...

donaldwj:
Can you tell me why the symlink command works? As far as i can tell it
should require a cd into the Run directory for the target to be valid.

donaldwj:
Apparently when target it not a full path it is treated as relative to
location of the created link. I had forgotten that 'ln -s' lets you create
links to things that may not exist yet which would be impossible if it
checked the target path at execution.

PR #327, opened 2019-07-10
Update Code of Conduct

mollymca:
Adjusted the text to comply with the UCAR contributor code of conduct https://www.ucar.edu/who-we-are/ethics-integrity/codes-conduct/contributors which is now official.


mollymca:
Sorry I missed the DOI mention. I tried to get the others. I copied directly from the website https://www.ucar.edu/who-we-are/ethics-integrity/codes-conduct/contributors . However, I could not find the DOI that is mentioned on this page. So, I would suggest removing that mention.

PR #329, opened 2019-07-11
Subsurface Routing Bugfixes

aubreyd:
This PR is based on the Subsurface Refactor candidate PR from James ( #315) and my Subsurface and Disag Pretty-ing Up PR (#328), so those should come first.

This PR fixes a few small bugs in the subsurface routing scheme. There are potential answer changes in anything impacted by soil moisture, particularly in flat areas. In a one-year NWM CONUS test, changes in streamflow were mostly minor (small peak reductions in flat areas). Impacts may be more significant in higher-res domains.

The code will now call the water table steepest head gradient neighbor check every LSM/subsurface routing timestep. This can slow model performance a bit. We are looking into better options for neighbor checks other than i,j loops, but for now this is functional. NWM CONUS performance was on average 4-5% slower with this fix.


aubreyd:
Results from CONUS experiment:
https://drive.google.com/open?id=1srdKNc3W0GpnK1DO08HD1puM2h3I0dLbVuBk_4p6uxA

aubreyd:
Thanks for the extra checks, Katelyn! I agree re a toy basin test. I have one I used for these checks and the variable soil depth work that might be a good candidate for this - smaller than Croton, useful for toy experiments where you need to see what every cell is doing, can manually manipulate values, and can easily do water budget checks, etc.

I'll check the conflicts against current master and see what is lingering...

kafitzgerald:
For reference, these are a few of the diffs from the nwm_ana test:

diff_zwattablrt
diff_sfcheadsubrt
diff_UGDRNOFF
diff_SFCRNOFF

aubreyd:
I pulled in latest master. There were conflicts, but in expected locations.

PR #330, opened 2019-07-21
Gridded channel slope fix

aubreyd:
Gridded channel slope check adjustment from Yates to fix instability in high flows in low slopes. Corrects negative flows in hyper-res and Luzon test cases. @dnyates and @laurareads should review/confirm I have this correct. Potentially answer changing (channel fluxes only) in gridded channel configs.


kafitzgerald:
So it sounds like this PR should go to v5.1.1 and then that will get merged to master.  @aubreyd wanted to do a bit more testing on a large(r) gridded domain first though.

aubreyd:
We are tabling this PR for now. Based on Irene testing, this change needs more investigation. Opening as a separate issue.

PR #331, opened 2019-07-22
Bucket parameter area normalization for NWM configuration

kafitzgerald:
This change adds basin area to the exponential bucket model removing the strong area dependence of this function and allowing for regionalization of parameters to catchments / buckets of differing size.

I had intended to put this in as an additional option for the bucket model via GWBASESWCRT along with a new formulation (I decoupled these changes and this will go in later), but some of the unlisted numbers are already in use, conceptually I think that this modification makes much more sense than the original implementation, and I'm nervous about the proliferation of options in the current context (it may make more sense post-refactor).  

Science results for this change have already been reviewed for CONUS as part of ongoing calibration work.


kafitzgerald:
Closing this and opening a PR w/ this and another formulation added as options.

PR #332, opened 2019-07-23
Additional bucket model options

kafitzgerald:
This PR adds two additional bucket model options (`GWBASESWCRT=4` and `GWBASESWCRT=5`) for the `UDMAP=1` (i.e. NWM) configuration of the model.  The former is a modification of the current `GWBASESWCRT=1` exponential model that normalizes to the area of the catchment / bucket so that parameters can be better regionalized and the latter is a new power law formulation.  

Namelist changes: new options for GW
Parameter file changes: none (both new options use the existing GWBUCKPARM.nc file and variables)


kafitzgerald:
Closing this.  Will open a separate PR for each option momentarily.

Issue #334, opened 2019-07-24
Excessive diurnal cycle in streamflow during spring snow melt season


barlage:
Model produces a streamflow diurnal cycle that has too high of an amplitude compared to observations in many spring snow melt dominated basins. A contribution to this high amplitude diurnal cycle is the way melt water is released from the snow pack. 

## Current Behavior
Currently, the model uses a parameter (SSI) to determine the fraction of the snow layer pore space that can be occupied by liquid water. Any liquid water above this amount drains to the layer below or if the bottom layer, drains from the pack to become infiltration or surface runoff. The ground surface acts as a barrier to this draining in the lowest model layer and should allow more liquid water to accumulate in the bottom of the snow pack

## Possible Solution
Do not limit the liquid water in the lowest snow layer. Instead, allow the liquid to accumulate at the base of the pack and set a time scale for release of this water from the pack based on liquid amount and a time scale parameter.

PR #335, opened 2019-07-25
Add a snow pack liquid water release timescale

barlage:
This PR addresses #334 

TYPE: enhancement

KEYWORDS: Noah-MP, snow model

SOURCE: Michael Barlage (NCAR)

DESCRIPTION OF CHANGES:

Do not limit the liquid water in the lowest snow layer. Instead, allow the liquid to accumulate at the base of the pack and set a time scale for release of this water from the pack based on liquid amount and a timescale parameter. The default timescale parameter SNOW_RET_FAC is set to 5e-5, which means 90% of the pack over-saturation (liquid in excess of porosity limit) drains over 12 hours when there is no new melt.

LIST OF MODIFIED FILES:
M module_sf_noahmplsm.F
M module_sf_noahmpdrv.F
M trunk/NDHMS/Land_models/NoahMP/run/MPTABLE.TBL
M  trunk/NDHMS/template/WCOSS/TBLS/MPTABLE.TBL

TESTS CONDUCTED:

  - Multi-year development tests in Taylor Park domain
  - One month CONUS NWM test performed by Aubrey


rcabell:
CONUS long-range regression has drift, `xrcmp` stats table below:

```
       Variable  Count     Sum      Min     Max   Range    Mean  StdDev  
model_discharge  4992  -2.510  -12.309  13.652  25.962  -0.001  0.961  
          nudge    48   4.318   -0.018   3.847   3.866   0.090  0.549  
```

kafitzgerald:
Do we not see this diff in other configurations?

rcabell:
The other configurations are still running, I will attach those results are they are available

barlage:
@rcabell Can you define drift? Does that mean the same as differences?

rcabell:
@barlage, yes, same as differences. My apologies for (unintentionally) introducing new nomenclature!

rcabell:
More complete output differences in the `nwm_long_range` configuration:

[PR335-nwm_long_range-diffs.txt](https://github.com/NCAR/wrf_hydro_nwm_public/files/3458742/PR335-nwm_long_range-diffs.txt)

rcabell:
Output differences in the `nwm_ana` configuration:

[PR335-nwm_ana-diffs.txt](https://github.com/NCAR/wrf_hydro_nwm_public/files/3461915/PR335-nwm_ana-diffs.txt)

rcabell:
The `channel_only` configuration passes regression over the CONUS domain

rcabell:
@barlage If you would please review the differences tables and see if they are acceptable, I will approve this PR

barlage:
@rcabell I'm not 100% on what these tables report but there is nothing that stands out as incorrect. 

Probably not the appropriate place for this but for future testing it might be good to output some plots from the testing, perhaps zoomed over some designated test basins. In a PR like this it would be very helpful since there should be no changes over basins that have no snow (or more specifically basins that have no snowmelt).

Issue #326, opened 2019-07-02
Level Pool using both inflow and previous timestep inflow


jdmattern-noaa:
The Level Pool physics code uses inflow from the current timestep and the previous timestep. It is unclear if the actual previous timestep inflow is passed to Level Pool.

My unit tests match a simulation output only when I pass the current inflow to both the current timestep inflow and the previous timestep inflow in Level Pool. Here is the call to Level Pool in Channel Routing:

https://github.com/NCAR/wrf_hydro_nwm_public/blob/master/trunk/NDHMS/Routing/module_channel_routing.F#L1899

Here is a comment in Channel Routing that seems to consider the current timestep inflow passed to Level Pool as the actual previous timestep inflow:

https://github.com/NCAR/wrf_hydro_nwm_public/blob/master/trunk/NDHMS/Routing/module_channel_routing.F#L1869

If this is the case, would the previous timestep inflow passed to Level Pools actually be from two timesteps back?

kafitzgerald:
@jdmattern-noaa was this fixed in one of your recent PRs?

jdmattern-noaa:
No. This is more of a question on our part for clarification. Nothing in my code would need to be updated based upon this.

Issue #312, opened 2019-06-07
Semantic bug in Levelpool


hellkite500:
A minor semantic bug in the over top logic.  A reservoir that is exactly full, should not be in the over top condition.  For most cases, this isn't noticeable, but it led to some strange (and incorrect) behavior when blending with the persistence module (IIRC certain conditions could allow a reservoir to release too little water.)  @jdmattern-noaa may be able to provide some more concrete examples.

Regardless, the semantic change to strictly greater than should be appropriate.
https://github.com/NCAR/wrf_hydro_nwm_public/blob/6170c18b591af21bdf97c75334deaba00833938f/trunk/NDHMS/Routing/Reservoirs/module_levelpool.F#L321

kafitzgerald:
It sounded like this was included in a different PR after #292 was closed.  I believe this was addressed in a recently merged PR. @jdmattern-noaa can you confirm?

jdmattern-noaa:
Yes, confirmed. The new free-board storage in Level Pool nullifies the
previous code and this issue.

PR #336, opened 2019-07-25
Add damlength var

rcabell:
Added support for new "Dam_Length" (in NetCDF) / DAML (Fortran) variable

Add additional Dam_Length parameter to the lakeparm.nc file for use in level_pool scheme. This was introduced to try to smooth out the response seen after implementation of #292.

_No issue currently exists to back this PR specifically (#312 and #326 are related); one should be opened and this description updated_


trunk/NDHMS/Data_Rec/module_RT_data.f90
```diff
@@ -0,0 +1,274 @@
...
+  REAL,  DIMENSION(50)     :: TOPWIDCC, NCC  !topwidth and mannings n of compund
+
+  ! VARIABLES FOR  RESERVOIRS
+  REAL, allocatable, DIMENSION(:)      :: RESHT  !reservoir height
+!-- lake params
+  INTEGER, allocatable, DIMENSION(:) :: LAKEIDA     !id of lakes in routlink file
+  INTEGER, allocatable, DIMENSION(:) :: LAKEIDM     !id of LAKES Modeled in LAKEPARM.nc or tbl
+  REAL, allocatable, DIMENSION(:)    :: HRZAREA    !horizontal extent of lake, km^2
+  REAL, allocatable, DIMENSION(:)    :: WEIRL      !overtop weir length (m)
+  REAL, allocatable, DIMENSION(:)    :: DAML      !overtop dam length (m)
```
hellkite500: If we are adding NEW variables that don't require a significant refactor throughout the code, can we try to move away from fixed format, F77 naming conventions?  DAM_LENGTH is much more descriptive, especially when seeing it in the code out of context with the associated comment.  New variables are the easiest to start with! And it looks like the addition of dam length in module_levelpool did this, so this name should be the same!

trunk/NDHMS/Routing/Reservoirs/module_levelpool.F
```diff
@@ -183,6 +185,7 @@ subroutine LEVELPOOL(ln,qi0,qi1,qo1,ql,dt,H,ar,we,maxh,wc,wl,oe,oc,oa)
     real, intent(IN)    :: we      ! bottom of weir elevation
     real, intent(IN)    :: wc      ! weir coeff.
     real, intent(IN)    :: wl      ! weir length (m)
+    real, intent(IN)    :: dl      ! dam length(m)
```
hellkite500: Once more, should be consistent with other names of this variable so that the intent and use of it is clear.  So far, three different variables names to describe the SAME thing, DAML, dam_length, dl.  They should all look the same, and let the encapsulating namespace help differentiate their context.

rcabell:
This PR is now passing all CONUS tests as well as the Travis CI test (due to updated Croton domain file). I recommend that it be merged. @kafitzgerald, @jmccreight  — if one (or both) of you second that, please mark the PR as approved and I will merge it. Thanks!

trunk/NDHMS/Data_Rec/module_RT_data.f90
junjie2008v: DAM_LENGTH=WEIRL ?
can I understand like that?

hellkite500: @junjie2008v I'm not sure I understand what your question is asking.  Is it related to this code review comment or just looking for clarity on the `DAM_LENGTH` variable in general?   If this is the case, I suggest opening an issue to start that conversation.   IIRC there might be a closed issue related to `DAM_LENGTH` that might be a good place to start as well.

junjie2008v: Thank you for taking the time out of your busy schedule to reply！&& What suggestions do you have for the DAM_LENGTH ?

PR #338, opened 2019-07-29
Groundwater option - exponential w/ area normalization

kafitzgerald:
This PR adds an additional bucket model option `GWBASESWCRT=4` for the `UDMAP=1` (i.e. NWM) configuration of the model per discussion / documentation in issue #354 . 

This formulation is a modification of the current `GWBASESWCRT=1` exponential model, but normalized to the area of the catchment / bucket so that parameters can be better regionalized.

**Namelist changes**: Additional option number available for `GWBASESWCRT`

**Parameter file changes**: No changes required

**Testing updates**: `GWBASESWCRT=4` will be the default option for NWM configuration tests

Test results from CONUS simulations are included via the link below.


kafitzgerald:
Here's a [link](https://docs.google.com/presentation/d/1K3snxzjUbWvyVnqXdOa7ofUeHvFr6vt7GEsD2ik5Ktw/edit?usp=sharing) with some plots from test simulations.

kafitzgerald:
CI failed as expected with diffs only in the channel, lakes, and bucket.

trunk/NDHMS/Routing/module_GW_baseflow.F
```diff
@@ -216,8 +216,14 @@ subroutine simp_gw_buck_nhd(           &
 
 ! Assume exponential relation between z/zmax and Q...
 !DJG force asymptote to zero to prevent 'overdraft'... 
-               qout_gwsubbas(bas) = C(bas)*(EXP(ex(bas)*z_gwsubbas(bas)/(z_mx(bas)/1000.))-1) !Exp.model. q_out (m^3/s)
+               if(GWBASESWCRT.eq.1) then ! Exponential model
+                   qout_gwsubbas(bas) = C(bas)*(EXP(ex(bas)*z_gwsubbas(bas)/(z_mx(bas)/1000.))-1) ! q_out (m^3/s)
+
+               elseif(GWBASESWCRT.eq.4) then ! Updated exponential model normalized by area
+                   qout_gwsubbas(bas) = C(bas)*basns_area(bas)*1.E-6*(EXP(ex(bas)*z_gwsubbas(bas)/(z_mx(bas)/1000.))-1) ! q_out (m^3/s)
```
jmccreight: I dont particularly like the magic number 1.E-6. I'm assuming that it's a km^2/m^2 conversion to convert basin area to meters. Would be more readable if it were a variable named to that effect.

jmccreight: Perhaps more mysterious is the preexisting 1000.

kafitzgerald: Yeah, both of these assumptions are true (km^2/m^2 and the units issue with respect to z).  

I can add these conversions in as variables quickly.  

I would have liked to have cleaned up this code a bit more, but was worried about timelines and wasn't sure if this was still on the docket to be refactored.

Issue #346, opened 2019-08-06
Issue examining land grid in module_NWM_io_dict.F


camelstation:
## Expected Behavior
Lines 1503-1521 of module_NWM_io_dict.F examine the land spatial metadata file (GEOGRID_LDASOUT_Spatial_Metadata.nc) to obtain the number of columns, etc. The metadata file is opened at line 1505 and grid attributes are extracted on the subsequent lines.

## Current Behavior
Line 1505 opens the metadata file and gives it the ncid of "ftnGeo". There are four subsequent lines that examine the attributes of this file (`iret = ....`), followed by it being closed line 1519. 

Two of these `iret = ....` use the ncid "ftnGeo", while the other two use "ftnMeta". As "ftnMeta" has not been defined, the model execution fails with: 

```
"The job is stopped due to the fatal error. 
ERROR: Unable to find x variable in LAND spatial metadata file"
```

## Possible Solution
As this section of code is opening and reading the metadata file, replace all "ftnGeo" with "ftnMeta". Example solution as follows

```
      iret = nf90_open(trim(nlst_rt(1)%land_spatial_meta_flnm),NF90_NOWRITE,ncid=ftnMeta)
      if(iret .ne. 0) then
         call postDiagMsg(diagFlag,'WARNING: Unable to open LAND Spatial Metadata file. Defaulting to missing values.')
         numColLand = -9999
         resLand = -9999
      else
         iret = nf90_inq_dimid(ftnMeta,'x',xDimId)
         call nwmCheck(diagFlag,iret,'ERROR: Unable to find x dimension in LAND spatial Metadata file')
         iret = nf90_inquire_dimension(ftnMeta,xDimId,len=numColLand)
         call nwmCheck(diagFlag,iret,'ERROR: Unable to retrieve number of columns in the LAND spatial metadata file')
         iret = nf90_inq_varid(ftnMeta,'x',xVarId)
         call nwmCheck(diagFlag,iret,'ERROR: CAMPBELL Unable to find x variable in LAND spatial metadata file')
         iret = nf90_get_att(ftnMeta,xVarId,'resolution',resLand) 
         call nwmCheck(diagFlag,iret,'ERROR: Unable to get x resolution in LAND spatial metadata file')
         iret = nf90_close(ftnMeta)
         call nwmCheck(diagFlag,iret,'ERROR: Unable to close the LAND spatial metadata file')
      endif

```

## Steps to Reproduce (for bugs)
[...]

PR #347, opened 2019-08-09
Bugfix - spatial metadata reads

kafitzgerald:
Fixes spatial metadata (LSM and Routing grid spatial metadata) file reads.

This should resolve issue #346 

Values computed from these sections (xRes and yRes) do not appear to be used elsewhere in the code, automated tests pass, and metadata in output files for RTOUT and CHRTOUT_GRID appear fine before and after the fix.


kafitzgerald:
This should probably go into v5.1.1 as well (it's a non-answer changing bugfix).  If others agree, I'll open a PR to that branch.

PR #357, opened 2019-09-03
Orchestrator config io

afanfa:
This PR includes the orchestrator layer, the configuration module and the IO layer.

PR #358, opened 2019-09-04
Noahmp jedi refactoring

afanfa:
Top level hrldas driver changes. It includes #357 and it will collapse after it gets merged.


jmccreight:
Addresses #360

PR #362, opened 2019-09-09
Xrcmp testing

jmccreight:
Needs https://github.com/NCAR/wrf_hydro_py/pull/190 to update wrfhydropy in order to pass testing. I think this wrfhydropy update also needs to get pushed to pypi.

1. The nan test was failing because of other reasons, syntax error, on the tests I was running. I'll note that running the tests with --pdb is sometimes very illuminating. I fixed that line [here](https://github.com/NCAR/wrf_hydro_nwm_public/pull/362/files#diff-6b294f935722038f503bf8abab66a9dfL16) 🤦‍♂ 

2. I found that xarray is SLOW for croton even in parallel. So I think we really should keep nccmp around. With nccmp around we dont have to deal with having 2 separate ways of doing metadata checks.

3. To invoke using xarray instead of nccmp (for data comparions only), the argument `--xrcmp_n_cores N` can be provided. Values greater than 1 mean "use xarray". Xrnan was refactored and sped up. It is always used for checking nans, but now it can do this in parallel, using the same `xrcmp_n_cores`.

4. I implemented the simplest thing I could in terms of `--use_existing_test_dir` it will simply skip compile or run when the target/run directories already exist for the test. The data comparisons always happen. This is a "power user" option. If you delete a run but not the compile, you will end up with the same run you deleted. But it will certainly mean that when reference is static, it only needs compiled and run once. It will be nice to script using this option.

5. Why is this in there? https://github.com/NCAR/wrf_hydro_nwm_public/pull/362/files#diff-75325934c6ce3d98d85edea4f0bad740


jmccreight:
Needs https://github.com/NCAR/wrf_hydro_py/pull/190 to update wrfhydropy in order to pass testing.

1. The nan test was failing because of other reasons, syntax error, on the tests I was running. I'll note that running the tests with --pdb is sometimes very illuminating. I fixed that line [here](https://github.com/NCAR/wrf_hydro_nwm_public/pull/362/files#diff-6b294f935722038f503bf8abab66a9dfL16) 🤦‍♂ 

2. I found that xarray is SLOW for croton even in parallel. So I think we really should keep nccmp around. With nccmp around we dont have to deal with having 2 separate ways of doing metadata checks.

3. To invoke using xarray instead of nccmp (for data comparions only), the argument `--xrcmp_n_cores N` can be provided. Values greater than 1 mean "use xarray". Xrnan was refactored and sped up. It is always used for checking nans, but now it can do this in parallel, using the same `xrcmp_n_cores`.

4. I implemented the simplest thing I could in terms of `--use_existing_test_dir` it will simply skip compile or run when the target/run directories already exist for the test. The data comparisons always happen. This is a "power user" option. If you delete a run but not the compile, you will end up with the same run you deleted. But it will certainly mean that when reference is static, it only needs compiled and run once. It will be nice to script using this option.

5. Why is this in there?

PR #365, opened 2019-09-13
Testing cleanups

jmccreight:
though the branch is xrcmp_testing, it really has little to do with xrcmp
mostly removing some unnecessary testing code, using `=` in the call to pytest to avoid future headaches, and improving readme files on testing


jmccreight:
Abusing my authority and merging this... it has no effect on the model. There's some possibility of a negative effect on testing edge cases, but that has very low risk.

PR #367, opened 2019-09-16
Bugfix - spatial metadata reads

kafitzgerald:
This is the fix from PR#347 (same commit).  It technically should have gone to v5.1.1 first as a bug fix and then have gotten merged to master.  My bad on that one.

From the previous PR...

> Fixes spatial metadata (LSM and Routing grid spatial metadata) file reads.
>
>This should resolve issue #346
>
>Values computed from these sections (xRes and yRes) do not appear to be used elsewhere in the code, automated tests pass, and metadata in output files for RTOUT and CHRTOUT_GRID appear fine before and after the fix.


rcabell:
@kafitzgerald the CI is showing that there's NaNs in the output. Is that expected?

kafitzgerald:
No.  I'll take a look.

kafitzgerald:
Just documenting here that the test failures are because of conflicts between the old tests and the CI environment.  Chatted with @rcabell and planning on back porting the updated tests since there are likely to be a few more PRs on this branch.

PR #380, opened 2019-10-03
Create PULL_REQUEST_TEMPLATE.md

jmccreight:


kafitzgerald:
Here's a link to the WRF pull request template that @rcabell was referring to: https://github.com/wrf-model/WRF/blob/master/.github/PULL_REQUEST_TEMPLATE

I'll let @rcabell add what he's interested in from there before approving, but I'm definitely on board with getting one of these in the repo as a guide for folks.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
+Use this template to give a detailed message describing the change you want to make to the code.
+If you are unclear on what should be written here, see  https://github.com/wrf-model/WRF/wiki/Making-a-good-pull-request-message for some general guidance
```
kafitzgerald: I wonder if we should copy this over as well and modify.  Probably not a bad idea to build out our wiki for some dev docs anyway 🤷‍♀ .

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+### Checklist
+Merging the PR depends on following checklist being completed. Add `X` between each of the square 
+brackets if they are completed in the PR itself. If a bullet is not relevant to you, please comment 
+on why below the bullet.
+
+ - [ ] Closes issue #xxxx (An issue must exist or be created to be closed. The issue describes and documents the problem and general solution, the PR describes the technical details of the solution.) 
+ - [ ] Tests added (unit tests and/or regression/integration tests)
+ - [ ] Backwards compatible
+ - [ ] Requires new files? If so, how to generate them.
+ - [ ] Fully documented
```
kafitzgerald: Ha, we need a better process for this or suggestions for the interim (maybe we do have some of these?).  

I'll see if I can get more stuff converted to markdown and in the repo soon.  Hopefully that makes things more manageable.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+ - [ ] Short description in the Development section of `NEWS.md`
```
kafitzgerald: Yes, please!

.github/PULL_REQUEST_TEMPLATE.md
rcabell: I agree that we should have our own wiki. If we don't want to delay this PR, though, we could just remove this line for now.

.github/PULL_REQUEST_TEMPLATE.md
rcabell: @jmccreight came up with these originally; perhaps he can weigh in with specific suggestions

.github/PULL_REQUEST_TEMPLATE.md
kafitzgerald: Sounds good to me.

PR #390, opened 2019-10-18
Hydro stop fix

donaldwj:
This pull request moves the function 'HYDRO_stop' from the file module_HYDRO_drv.F to utils/module_hydro_stop.F. The function is now part of the module module_hydro_stop where previously it was a global subroutine. These changes simplify the creation of testing code for sub-section of the model which would other wise need to either compile module_HYDRO_drv.F or create alternative definitions of the subroutine.


trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
```diff
@@ -1816,82 +1816,42 @@ end subroutine lsm_input
 end module module_HYDRO_drv
 
 ! stop the job due to the fatal error.
```
kafitzgerald: Are we missing some use statements now that this has been moved out from module_HYDRO_drv.F?

donaldwj: Any place that calls hydro_stop() should have the required use statements added by this pull request.

rcabell: `module_HYDRO_drv.F` doesn't have a use statement for it but it does compile... 

There must be a nested `use` that brings it into the namespace, but it's not immediately obvious where it's coming from. In any case, yes, any place that calls `hydro_stop()` should have an _explicit_ `use` statement

donaldwj:
Admittedly I only added them where the lack was generating missing symbols.
I go and add the declaration to other files.

rcabell:
@donaldwj: Great, thanks!

donaldwj:
The list of files that needed to be updated could be found with this commands 'grep -L "use module_hydro_stop" $(grep -rinl "call hydro_stop" .)'

PR #419, opened 2019-11-07
Add NWM_META compile time option

### LINKED ISSUE #418: NCEP WCOSS compile time option does too much?
## Current Behavior
The NCEP_WCOSS flag impacts output file metadata in addition to handling other system specific issues making testing and prototyping on other systems difficult.

## Possible Solution
Split the NCEP_WCOSS flag into two.  One for the WCOSS file units, etc. that are specific to the system and one (e.g. NWM_META) for NWM specific IO metadata.

kafitzgerald:
TYPE: enhancement

KEYWORDS: compile option, NWM, metadata

SOURCE: Katelyn (NCAR)

DESCRIPTION OF CHANGES: 
* Adds a new compile time option NWM_META to control NWM specific metadata rather than using NCEP_WCOSS for this in addition to its other functionality

ISSUE:
```
Fixes #418
```

TESTS CONDUCTED: 
Local test suite passes

NOTES: 
Open to other ideas for how to address this issue.  I just needed something for the time being and this worked.


rcabell:
Just to clarify, will this require NCO to set both `NCEP_WCOSS` _and_ `NWM_META` when they do an operational build? If so, that should probably be documented somewhere (unless I'm not seeing that it already has been)

kafitzgerald:
It will require both to be set.  I thought about setting NWM_META along with NCEP_WCOSS, but opted not to since it seemed cleaner this way.  Open to thoughts on that front.  

I'll see if I can track down the instructions for NCO and/or log that somewhere in here.

rcabell:
This PR passes AnA/Long-Range Automated CONUS Testing on Cheyenne. Recommend merging.

Issue #230, opened 2018-11-29
Default to netCDF large file support to be consistent with WRF


kafitzgerald:
## Expected Behavior
Given that our suggestion to users is to compile with netCDF large file support and WRF now has that as their default, it seems like it might be worth changing this is our code base.  

## Current Behavior
When WRFIO_NCD_LARGE_FILE_SUPPORT is not set or set to 0 we compile w/out large file support 

## Possible Solution
The solution they implemented [here](https://github.com/wrf-model/WRF/pull/112) could work for us (i.e. removing WRFIO_NCD_LARGE_FILE_SUPPORT and implementing a new option WRFIO_NCD_NO_LARGE_FILE_SUPPORT to allow users to explicitly turn large file support off.

Benefits:
* This would make our default consistent with what we recommend to users
* Consistency with WRF allowing us to (hopefully) just drop in new versions of our code

kafitzgerald:
Reopening as a reminder to add this to master as well.

PR #426, opened 2019-11-13
V5.1.1 IO update

kafitzgerald:
TYPE: enhancement?

KEYWORDS: IO, netCDF, compile option, WRF

SOURCE: Katelyn (NCAR)

DESCRIPTION OF CHANGES: 
* Removes `WRFIO_NCD_LARGE_FILE_SUPPORT` compile time option.  This is no longer the WRF default (it's been replaced with a flag that does the opposite).  
* Rather than implement this other flag these netCDF3 outputs were updated to netCDF4, which is already required elsewhere.
* Updates `nf_create` function to `nf90_create`
* Changes deprecated `NF90_HDF5` mode to `NF90_NETCDF4` 

ISSUE: Addresses issue #230

TESTS CONDUCTED: Test suite via testing container.  **Fails testing where netCDF file type has changed.**

NOTES: 
* Relevant to v5.1.1 (for integrating with WRF) and master
* No huge rush to merge this (other than the pending community release)
* May want to consider other approaches (e.g. implementing the same flag as WRF and maintaining some netCDF3 outputs)
* We're using the netCDF4 output format (already in some places and in more after this PR), but as far as I can tell don't require features from the [netCDF4 enhanced data model](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_model.html).  Given the state of support for features in the enhanced data model (not great) in various analysis + display tools, we might consider setting the `NF90_CLASSIC_MODEL` flag to enforce the classic data model.  

### Checklist

 - [x] Closes issue #230
 - [-] Tests added (unit tests and/or regression/integration tests)
 - [x] Backwards compatible
 - [-] Requires new files? No
 - [-] Fully documented
 - [-] Short description in the Development section of `NEWS.md`


hellkite500:
"We're using the netCDF4 output format (already in some places and in more after this PR), but as far as I can tell don't require features from the netCDF4 enhanced data model. Given the state of support for features in the enhanced data model (not great) in various analysis + display tools, we might consider setting the NF90_CLASSIC_MODEL flag to enforce the classic data model."

The enhanced data model is backwards compatible, and tooling IS adding support for new features (some faster than others).  Also, with older versions of the netcdf C library, issues exist in working with the classic model on some tools (see https://github.com/pydata/xarray/issues/2822, for example).  I see no reason to strictly enforce the classic model, when it is a compatible subset of the enhanced.  Also, adopting the enhanced model everywhere makes adopting new features simpler in the future.

kafitzgerald:
I didn't implement this anyway, but for what it's worth I wasn't talking about going back to netCDF3 just restricting to the features of the classic model for the sake of compatibility.  Per the docs:

>Oring the NF90_CLASSIC_MODEL flag with the NF90_NETCDF4 flag causes the resulting netCDF-4/HDF5 file to restrict itself to the classic model - none of the new netCDF-4 data model features, such as groups or user-defined types, are allowed in such a file.

This could easily be changed when there's broader support for the enhanced features.  

That said, I'm fine with leaving it is as.

rcabell:
I like that this removes `WRFIO_NCD_LARGE_FILE_SUPPORT` -- we should add ("add"? ) that to the `master` branch as well...

rcabell:
This PR passes automated CONUS regression tests for AnA and Long-Range with the expected exception of metadata. The following are the testing-related message for this changes:

```
NC_FORMAT_64BIT <> NC_FORMAT_NETCDF4
```

```
/glade/u/home/rcabell/.local/lib/python3.6/site-packages/wrfhydropy/core/simulation.py:263: UserWarning: Model minor versions v5.1.1-beta
   do not match domain minor versions v5.1.0-beta2
  
    domain.compatible_version)
```

The issue with the domain version should be addressed by another Issue/PR and need not be changed here.

PR #452, opened 2020-02-12
bug fix - domain decomposition

### LINKED ISSUE #444: MPI process topology should respect grid dimensions
## Expected Behavior
Determination of nx and ny for processes should be informed by grid dimensions

## Current Behavior
This is calculated in the getNX_NY subroutine in mpp_land.F without knowledge of the land surface model (or any) grid dimensions.

For small and/or vectorized domains this can result on topologies that don't make sense for the grid and at a minimum causes problems in associating land grid cells with catchments / reached in the user defined mapping code.  This results in simulations with no streamflow and no errors are thrown.

## Possible Solution
Add some logic and error checking to the subroutines for specifying process topology.

We may also want to add some error checking in the relevant UDMP subroutine(s).

## Steps to Reproduce (for bugs)
Run the code with @aubreyd's dummy vectorized domain and np>1.

kafitzgerald:
TYPE: bug fix

KEYWORDS: MPI, domain decomposition

SOURCE: Katelyn, NCAR

DESCRIPTION OF CHANGES: 
Determination of nx and ny for MPI processes should be informed by grid dimensions.  This was not the case previously and is problematic for very small and/or vectorized domains.  

The grid dimensions of the land surface model were added to MPP_LAND_INIT() as arguments, additional logic added to the grid decomposition, and call(s) adjusted accordingly.

Open to other ideas for how to address this.  I'm not super familiar w/ these parts of the code and we just needed something to get this working for a project.

ISSUE: Fixes #444

TESTS CONDUCTED: Test runs on a 1D test domain.  

NOTES: 
* Only implemented for NoahMP (Noah implementation would require moving more things around in the initialization and we should probably implement the orchestrator at the same time).  
* Should be checked to make sure it doesn't interfere w/ coupled parallelization.
* May want to add additional error checking.

### DIFF
--- a/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
+++ b/trunk/NDHMS/HYDRO_drv/module_HYDRO_drv.F
@@ -1369,9 +1369,6 @@ integer:: ix0,jx0
 integer, dimension(ix0,jx0),optional :: vegtyp, soltyp
 integer            :: iret, ncid, ascIndId
 
-#ifdef MPP_LAND
-call  MPP_LAND_INIT()
-#endif
 
 
 ! read the namelist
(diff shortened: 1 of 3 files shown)


rcabell (approved):
PR passes CONUS-CI testing with no diffs.

rcabell (approved):

PR #472, opened 2020-03-10
Minor change to remove command to copy non-existent tables

kafitzgerald:
TYPE: cleanup

KEYWORDS: build, tables, nwm

SOURCE: Katelyn, NCAR

DESCRIPTION OF CHANGES: 
Minor change to avoid warning about tables for NWM build.  I'd considered adding these back in, but it seems best to leave them with the other materials and create a clean run directory.


kafitzgerald:
@rcabell this should address the edge case where we had CI failures now.

PR #480, opened 2020-04-22
Improving performance of nudging routines

afanfa:
TYPE: enhancement

KEYWORDS: performance, nudging, better data layout

SOURCE: Alessandro Fanfarillo, NCAR

DESCRIPTION OF CHANGES: The observations used in the nudging routines are organized as Array of Structures. This layout penalizes performance in many ways. This PR reorganize the layout to be a Structure of Arrays.

ISSUE: None

TESTS CONDUCTED: All the tests pass on Cheyenne except reach! According to @jmccreight reach shouldn't even be using nudging. I am creating the PR to test reach using TravisCI.

NOTES: The PR requires some cleanup.

### Checklist
Merging the PR depends on following checklist being completed. Add `X` between each of the square 
brackets if they are completed in the PR itself. If a bullet is not relevant to you, please comment 
on why below the bullet.

 - [ ] Closes issue #xxxx (An issue must exist or be created to be closed. The issue describes and documents the problem and general solution, the PR describes the technical details of the solution.) 
 - [ ] Tests added (unit tests and/or regression/integration tests)
 - [ ] Backwards compatible
 - [ ] Requires new files? If so, how to generate them.
 - [ ] Fully documented
 - [ ] Short description in the Development section of `NEWS.md`

PR #487, opened 2020-04-30
Performance improvements in initialization routines

afanfa:
TYPE: enhancement

KEYWORDS: performance improvement, hash to build topology, initialization

SOURCE: Alessandro Fanfarillo - NCAR

DESCRIPTION OF CHANGES: Several routines during the initialization were implemented with loops over the arrays of local and global links. This PR replaces those loops with searches into a hashtable-like structure (C++ Map from STL). This new approach is faster than the previous one and I can be improved even more by using a real hashtable instead of a C++ Map data structure.
The use of C++ is due to convenience more than functionality. The solution proposed in this PR is pretty general and it can be easily implemented in any language.

ISSUE: None

TESTS CONDUCTED: nwm_ana, nwm_long_range, gridded and reach using the croton_ny domain.

NOTES: A new file called hash_routines.cc has been added into utils. For readability reasons, the code replaced by the new routines has been added as a comment in the code right before the function invocation.

### Checklist
Merging the PR depends on following checklist being completed. Add `X` between each of the square 
brackets if they are completed in the PR itself. If a bullet is not relevant to you, please comment 
on why below the bullet.

 - [ ] Closes issue #xxxx (An issue must exist or be created to be closed. The issue describes and documents the problem and general solution, the PR describes the technical details of the solution.) 
 - [X ] Backwards compatible
 - [X ] Requires new files? If so, how to generate them.
 - [X ] Fully documented


rcabell:
Couple of things: 

* can you rebase this PR to reduce the number of commits to a few small, functional ones? If not, I can just do a squash merge but the history will be lost

* I'm not sure we want to introduce a dependency on C++ (or, really, any other language) at this time. How hard would it be to code the hashmap in Fortran? I found `https://github.com/jl2922/fhash` that might be a good starting point.

rcabell:
Also, this is answer-changing for channel routing, so we'll need to identify if that's expected (and/or acceptable)

afanfa:
Yes, on Cheyenne I kept getting the same errors I got for #480 but apparently this time it was right. I can't reproduce the same problem identified by TravisCI (on Cheyenne the differences are NaNs, on travisCI are actual numbers).
About the commits, I am not sure I can do a rebase. There is a little in most of them...
About the dependency on C++ I totally understand. I didn't know what was better, C++ dependency or external library dependency (like hash). If the latter is more acceptable I am totally open to it.

afanfa:
Now that all tests pass, I am going to implement a simple hashtable in Fortran and replace it with the calls to the C++ routine. Considering the amount of work required and the current number of commits, I am going to close this PR and open a new one with fewer commits.

PR #502, opened 2020-06-16
PGI Fortran 20.4 compatibility patches

### LINKED ISSUE #460: Build fails using PGI compilers
This is somewhat of a known issue... but figured it was worth documenting if nothing else as a reminder to remove PGI from the list of supported compilers for next release and perhaps hide that build option in the configure script (it is #1 after all).

## Current Behavior
Builds using PGI compilers fail because of modern Fortran features

## Possible Solution
Remove statements of support for PGI from docs / FAQs

## Steps to Reproduce (for bugs)
Attempt a build w/ env listed below

## Your Environment
* Version of the code used: current master
* Operating System and version: Cheyenne
* Compiler and version: PGI 19.10
* Other relevant information: v5.1.x seems to build fine w/ PGI 19 compilers

rcabell:
TYPE: compatibility enhancement

KEYWORDS: compiler

SOURCE: Ryan@NCAR

DESCRIPTION OF CHANGES: PGI Fortran version 20.4 finally supports the Fortran 2008 features used in the model, but a few small changes were required to build correctly and run without crashing when optimization are turned on. Specifically, some variables in the `readBucket_nhd` subroutine  needed to be marked as `volatile` so the PGI optimizer wouldn't move them out of loops. Also, PGI doesn't seem to like static assignment of type-bound procedure pointers to netCDF NF90 routines, so these assignments were moved to the IO_Manager initializer. 

ISSUE: 
```
Fixes #460 
```

TESTS CONDUCTED: Compared values in Croton with Intel build and they match within 1E-4 places. 

**NOTE**: The model built with PGI Fortran, even with optimizations turned on, is 15-20% slower than the Intel Fortran build, so for performance reasons this compiler is not recommended unless no better compiler is available.

### Checklist
Merging the PR depends on following checklist being completed. Add `X` between each of the square brackets if they are completed in the PR itself. If a bullet is not relevant to you, please comment on why below the bullet.

 - [X] Closes issue #460 (An issue must exist or be created to be closed. The issue describes and documents the problem and general solution, the PR describes the technical details of the solution.) 
 - [X] Backwards compatible
 - [ ] Requires new files? If so, how to generate them.
 - [ ] Fully documented

### DIFF
--- a/trunk/NDHMS/Routing/module_HYDRO_io.F
+++ b/trunk/NDHMS/Routing/module_HYDRO_io.F
@@ -9317,19 +9317,20 @@ end function get_netcdf_dim
 subroutine readBucket_nhd(infile, numbasns, gw_buck_coeff, gw_buck_exp, &
                 gw_buck_loss, z_max, z_init, LINKID, nhdBuckMask)
     implicit none
-    integer :: numbasns
+    integer, intent(in) :: numbasns
     integer(kind=int64), dimension(numbasns) :: LINKID
     real, dimension(numbasns) :: gw_buck_coeff, gw_buck_exp, gw_buck_loss
     real, dimension(numbasns) :: z_max, z_init
     integer, dimension(numbasns) :: nhdBuckMask
-    character(len=*) :: infile
+    character(len=*), intent(in) :: infile
 !   define temp array
-    integer :: i,j,k, gnid, ncid, varid, ierr, dimid, iret
+    integer, volatile :: i,j,k, gnid, ncid, varid, ierr, dimid, iret
     integer(kind=int64), allocatable, dimension(:) :: tmpLinkid
     real, allocatable, dimension(:) :: tmpCoeff, tmpExp, tmpLoss
     real, allocatable, dimension(:) :: tmpz_max, tmpz_init
 
 !   get gnid
+    gnid = 0
 #ifdef MPP_LAND
     if(my_id .eq. io_id ) then
 #endif
@@ -9343,7 +9344,7 @@ subroutine readBucket_nhd(infile, numbasns, gw_buck_coeff, gw_buck_exp, &
                !print*, "nf90_inq_dimid:  BasinDim"
                call hydro_stop("Failed read GBUCKETPARM - nf90_inq_dimid:  BasinDim")
        endif
-       iret = nf90_inquire_dimension(ncid, dimid, len = gnid)
+       iret = nf90_inquire_dimension(ncid, dimid, len=gnid)
     endif
     call mpp_land_bcast_int1(gnid)
 #endif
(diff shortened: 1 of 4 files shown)


kafitzgerald (dismissed):
I'll let Alessandro comment on the changes to type-bound procedure pointers, but looks good to me otherwise.

afanfa:
I don't love this approach, but I understand that having compatibility is important too. I just had a chat with another Fortran enthusiast and I might have a solution that removes the procedure pointers. I may have a solution by the end of this week. If you are in a rush to merge this PR just let me know, I can accept this PR and create a new issue with the new changes.

rcabell:
> I don't love this approach, but I understand that having compatibility is important too. I just had a chat with another Fortran enthusiast and I might have a solution that removes the procedure pointers. I may have a solution by the end of this week. If you are in a rush to merge this PR just let me know, I can accept this PR and create a new issue with the new changes.

Absolutely no rush to merge this, and I agree that the approach isn't ideal. If you have another solution, I'll wait to see it.

trunk/NDHMS/Routing/module_HYDRO_io.F
```diff
@@ -9317,19 +9317,21 @@ end function get_netcdf_dim
...
+    integer, intent(in) :: numbasns
     integer(kind=int64), dimension(numbasns) :: LINKID
     real, dimension(numbasns) :: gw_buck_coeff, gw_buck_exp, gw_buck_loss
     real, dimension(numbasns) :: z_max, z_init
     integer, dimension(numbasns) :: nhdBuckMask
-    character(len=*) :: infile
+    character(len=*), intent(in) :: infile
 !   define temp array
-    integer :: i,j,k, gnid, ncid, varid, ierr, dimid, iret
+    integer, volatile :: i,k
```
scrasmussen, line 9327: I just realized we can change these integers to volatile for only PGI compilers by sneaking in a simple macro. Helps with portability. For instance
```
integer &
#ifdef __PGI
   , volatile &
#endif
      :: i,k, gnid, ncid, varid, dimid, ierr, iret
```

rcabell: I've found that Cray Fortran often needs the `volatile` attribute as well -- is there any harm to having these here for all compilers?

> I just realized we can change these integers to volatile for only PGI compilers by sneaking in a simple macro. Helps with portability. For instance
> 
> ```
> integer &
> #ifdef __PGI
>    , volatile &
> #endif
>       :: i,k, gnid, ncid, varid, dimid, ierr, iret
> ```

scrasmussen (approved):
Everything looks good to me!

scrasmussen (approved):

PR #510, opened 2020-08-28
move WPC and AREAC initialization into the loop

jameshalgren:
The second secant calculation frequently receives a garbage value from the previous calculation.
It does not appear that this is by design. Making the initialization within the loop prevents the accidental value passing. This *changes the output* of the Muskingum Cunge calculation in many cases.

@donaldwj had separated the two secant calculations into separate subroutines in his refactored version of this code. I would suggest that we make that change as well, as it is much easier to see what the scope of the calculations should be. 

@kumdonoaa - thanks for your sharp eyes
@groutr - we'll need to update this in the new routing repository


jameshalgren:
@laurareads @dnyates 
I can provide a test python script that runs two versions of the SUBMUSKINGCUNGE module side by side.  Here is some example output which shows the unexpected propagation of values and the resulting difference in function output. 
``` 
$ ipython mc_WPC_AREAC_demo.py
['f2py3', '-c', 'varPrecision.f90', 'MUSKINGCUNGE.f90', '-m', 'mc_wrf_hydro']
['f2py3', '-c', 'varPrecision.f90', 'MUSKINGCUNGE_New.f90', '-m', 'mc_wrf_hydro_new']
running model with out-of-loop initialization (the way we do it now...)
 bfd:   28.5599995
 secant1 --WP:   215.813599     WPC:   0.00000000
 secant2 --WP:   259.508301     WPC:   646.299988
 secant1 --WP:   259.508301     WPC:   646.299988
 secant2 --WP:   203.432831     WPC:   646.299988
computed q: 46471.5078125 vel: 18.03421974182129 depth: 17.702878952026367
running model with in-the-loop initialization (the 'new' way...)
 bfd:   28.5599995
 secant1 --WP:   215.813599     WPC:   0.00000000
 secant2 --WP:   259.508301     WPC:   646.299988
 secant1 --WP:   259.508301     WPC:   646.299988
 secant2 --WP:   203.432831     WPC:   0.00000000
 secant1 --WP:   203.432831     WPC:   0.00000000
 secant2 --WP:   201.416199     WPC:   0.00000000
computed q: 46502.3515625 vel: 17.765806198120117 depth: 17.226545333862305
 ```

jameshalgren:
@jdhondia-noaa @jhrehanoaa -- take note. We'll probably need to make a few changes in response to this.

jmccreight (dismissed):
I believe this is a correction or a bugfix. I reviewed with @jameshalgren and @laurareads. 
we are not convinced that it is worth undrestanding the changes to the streamflow. if it is easy to do this with the CONUS-48h run, then we might consider comparing the results. But differences would depend on if the compound channel is activated.

we also confirmed that a11c3b0 appears to fail for the right reasons

rcabell:
I've started a CONUS-48 test; if the results are unacceptable, we will back this PR out.

jameshalgren:
Per our review, clarified the comment above to show the inconsistency with depth below bankful with a persistent WPC (compound channel Wetted Perimeter) value. 

> I believe this is a correction or a bugfix. I reviewed with @jameshalgren and @laurareads.
> we are not convinced that it is worth undrestanding the changes to the streamflow. if it is easy to do this with the CONUS-48h run, then we might consider comparing the results. But differences would depend on if the compound channel is activated.
> 
> we also confirmed that a11c3b0 appears to fail for the right reasons

laurareads:
I approve this for merging into the code.

jameshalgren:
Is there summary (or perhaps selected) output from the prior/new CONUS-48 tests we could post with the PR comments here to nominally document the impact?

kafitzgerald:
Noting that there's some question about what the correct target branch / release should be for this.

rcabell:
> Noting that there's some question about what the correct target branch / release should be for this.

It seems high-risk to put it into `5.2.0` vs `master` but I'll wait to hear the final consensus.

laurareads:
I have completed a channel-only CONUS test during Hurricane Irene, comparing NWMv2.1 and PR-510 fix for several gauges that saw high flows during the event.  I am finding very little difference in the streamflow between these two simulations.  I will conclude that the impact is minimal, at least for the test conducted.  My vote is to go ahead and put this into v2.1 if possible.   See attached hydrographs for the results of this test.
[navy_plots.zip](https://github.com/NCAR/wrf_hydro_nwm_public/files/5171032/navy_plots.zip)

jameshalgren:
Rebased to latest master.

kafitzgerald (approved):

rcabell (approved):

jameshalgren:
Thanks!

Issue #521, opened 2020-10-26
Incorrect namelist check for lakes physics options


kafitzgerald:
It looks like the namelist check for lake physics was reverted to the older version when the orchestrator was implemented disallowing reach lakes.  Should be an easy fix.

rcabell:
Closed by PR #522

PR #522, opened 2020-10-26
Remove no longer relevant namelist check for reach lakes

kafitzgerald:
Addresses issue #521 by removing an old namelist check that was added back in with the orchestrator.


rcabell (approved):

donaldwj (commented):
Fairly simple if this check is no longer needed then this is fine.

donaldwj (approved):
Fairly simple if this check is no longer needed then this is fine.

PR #524, opened 2020-10-26
Add -std=legacy to macros to support gnu >= 10.0.1

### LINKED ISSUE #523: Build fails with gfortran >= 10.0.1
## Expected Behavior
WRF-Hydro model should compile with gfortran 10 similarly to gfortran 9 and without errror

## Current Behavior
Errors occur on some files when building with current options:

`Error: Rank mismatch between actual argument at (1) and actual argument at (2) (scalar and rank-1)`

## Possible Solution
Adding "-fallow-argument-mismatch" to the macros.mpp.gfort file allows the code to build with gfortran-10, but breaks gfortran-9. Using "-std=legacy" allows the code to build with both versions but may have unforeseen side-effects.

rcabell:
TYPE: enhancement

KEYWORDS: gfortran, gnu

SOURCE: Ryan@NCAR

DESCRIPTION OF CHANGES: Add -std=legacy flag to macros.mpp.gfort to support gfortran >= 10.0.1

ISSUE: 
```
Fixes #523 "Build fails with gfortran >= 10.0.1"
```

### DIFF
--- a/trunk/NDHMS/arc/macros.mpp.gfort
+++ b/trunk/NDHMS/arc/macros.mpp.gfort
@@ -55,13 +55,13 @@ endif
 
 
 RMD		=	rm -f
-COMPILER90=	mpif90  
-F90FLAGS  =       -w -c -ffree-form -ffree-line-length-none -fconvert=big-endian -frecord-marker=4 
-MODFLAG	=	-I"./" -I"../../MPP" -I"../MPP" -I"../mod"
-LDFLAGS	=	
-CPPINVOKE	=   -cpp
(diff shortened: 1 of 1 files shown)


kafitzgerald (commented):
Should this go to v5.2.x instead of master?  Otherwise looks good

rcabell:
> Should this go to v5.2.x instead of master? Otherwise looks good

It'll go in both when it's finally approved.

kafitzgerald (dismissed):

kafitzgerald (approved):

donaldwj (approved):

PR #535, opened 2021-01-29
Update master with v5.2.x

rcabell:


rcabell:
Updated using an inverse-rebase from the command line, since we can't do that from GitHub

PR #550, opened 2021-04-28
Fix NUOPC cap for GNU compilers

danrosen25:
TYPE: fix

KEYWORDS: NUOPC, GNU

SOURCE: Daniel Rosen (drosen@ucar.edu) - NCAR/ESMF

DESCRIPTION OF CHANGES: Fix GNU compiling issues

ISSUE: `none`

TESTS CONDUCTED: Explicitly state if regression, integration, or unit tests were run before submitting (do not include the CI tests automatically run by GitHub)

### DIFF
--- a/trunk/NDHMS/CPL/NUOPC_cpl/WRFHydro_NUOPC_Cap.F90
+++ b/trunk/NDHMS/CPL/NUOPC_cpl/WRFHydro_NUOPC_Cap.F90
@@ -365,15 +365,15 @@ module WRFHydro_NUOPC
       defaultValue="0", convention="NUOPC", purpose="Instance", rc=rc)
     if (ESMF_STDERRORCHECK(rc)) return  ! bail out
     diagnostic = ESMF_UtilString2Int(value, &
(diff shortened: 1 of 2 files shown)


danrosen25:
This PR is on hold. I'm seeing bad values in the results using GNU 9.1.0. I've also had some trouble with variable initialization that may be due to compiler optimization. I may try to run with reduced optimization or switch the compiler when I have more time.

scrasmussen:
> This PR is on hold. I'm seeing bad values in the results using GNU 9.1.0. I've also had some trouble with variable initialization that may be due to compiler optimization. I may try to run with reduced optimization or switch the compiler when I have more time.

Hi @danrosen25, had a PR meeting and was just wondering if the status of this PR had changed at all? Is there a test case on Cheyenne that you could point us to?

 I had tested this PR, but not with NUOCP coupling... which kinda seems like the import part ; )

danrosen25:
Hi @scrasmussen, this is still on hold I haven't been able to go through the debugging process to determine why the different compiler is producing different results. I'll have to run each component individually using Intel and GCC and see which component is different.

PR #573, opened 2021-07-21
Change snow NIR albedo to use correct BATS NIR parameter.

aubreyd:
TYPE: bug fix

KEYWORDS: snow, albedo, parameters

SOURCE: Aubrey D, NCAR 

DESCRIPTION OF CHANGES: Fix what I think is a typo in the NIR snow albedo calculation. Should not result in answer changes if using default parameters (since BATS_VIS_DIR = BATS_NIR_DIR in default).

TESTS CONDUCTED: Sample basin tested to confirm no answer changes with default params and answer changes when BATS_VIS_DIR != BATS_NIR_DIR


aubreyd:
@cenlinhe agrees with this fix

rcabell (approved):
Regression tests pass, and code change has been verified against NoahMP dev. Looks good to me.

PR #574, opened 2021-07-22
NCO Bugzilla [Bug 1164] 

zhengtaocui:
TYPE: bug fix

KEYWORDS: NCO Bugzilla 1164, Array out of bound error,  'temporary arrays are created' warnings

SOURCE: Zhengtao Cui, NOAA NWS OWP 

DESCRIPTION OF CHANGES: 
    NCO requested Bugzilla tickets to be fixed.

    [Bug 1164] New: Array out of boundary in nwm.fd, During NWM.v2.1 IT test, job nwm_analysis_assim failed at executable nwm.exe with following message when compiled with "-check all" option: forrtl: severe (408): fort: (3): Subscript added software requirements #1 of the array CH_LNKRT has value 0 which is less than the lower bound of 1.

     The variable new_start_i in module_RT.F starts from 0, however, the array CH_LNKRT is defined to start from 1. Changed variable new_start_i to start from 1 instead of 0.
     
     See the NOTE section about the warning messages that need to be fixed.

ISSUE: Fixes #574 

TESTS CONDUCTED: Tested on Hawaii domain. No changes of answers were found.

NOTES: 
   There is another issue to be addressed. In addition to the 'out-of-bounds' crash when 'check all' is enabled, the program also prints a huge amount of warning messages about "forrtl: warning (406): fort: (1): In call to MPI_SEND, an array temporary was created for argument #1". The log file has a size of 3.3G. This message came from the subroutine decompose_data_real in MPP/mpp_land.F. So it looks like it is related to the domain decomposition algorithm.   

  The ifort document says that the warning message is because the data sent by MPI_SEND is not in a contiguous block of memory. The compiler creates a temporary block of contiguous memory. MPI_SEND is not the only one that has this warning. Another one is the GETSUB8 function. So it appears it is something that in the algorithm used to send data between the processors, such as domain decomposition.


rcabell:
There are a considerable number of extraneous debug print statements in this commit that I'm concerned will make the logs extremely large when the HYDRO_D flag is enabled for science/physics development. If there are some that you believe would be useful to other developers, feel free to leave them in, but please remove the rest. Thanks!

zhengtaocui:
Thanks for the comments. I've removed the extraneous debug prints in HYDRO_D. Please let me know if anything else I can do.

scrasmussen (approved):
Changes look good to me. Instead of sending a slice of an array as a function argument, I created a one-dimensional array to hold that data. Passing this new one-dimensional array as the function argument fixes the huge amount of 'temporary array` warnings that we were getting when run with the check-all flag. I compared the output from Croton, NY with the master branch and the results were identical.

rcabell (approved):

PR #584, opened 2021-10-04
Exposed NoahMP swe limit parameter and snow bug fix for large packs

aubreyd:
TYPE: bug fix, enhancement

KEYWORDS: NoahMP, snow, parameters

SOURCE: Aubrey D, NCAR

DESCRIPTION OF CHANGES: 
1) NoahMP snow model had a hard-coded SWE cap (initially 2m, upped to 5m). I just exposed this cap to a user-defined parameter in MPTABLE global parameter section.
2) Fix for energy NAs over large snowpacks due to very small numbers being rounded to 0, so number/number=NA instead of 1. Switched to an if statement.

TESTS CONDUCTED: Tested with NWM-Alaska domain. Limit value change experiments behave as expected. Bug fix also tested with NWM-Alaska.

NOT BACKWARD COMPATIBLE! Requires updated MPTABLE. Code should crash if new parameter not present.


rcabell (approved):
Tests failed, but only due to: 

```DIFFER : VARIABLE : SNEQV : ATTRIBUTE : valid_range : VALUES : 0,100000 <> 0,1000000```

Which is expected, and an intended effect of this PR.

scrasmussen (approved):
LGTM!

rcabell (approved):
Only test failure is the expected `SNEQV : ATTRIBUTE : valid_range : VALUES : 0,100000 <> 0,1000000` on both Croton and CI. Recommend merge.

PR #602, opened 2022-02-09
Update WRF coupler to recent MPP_LAND_INIT changes

rcabell:
TYPE: bug fix

KEYWORDS: coupled, MPP

SOURCE: Ryan @ NCAR

DESCRIPTION OF CHANGES: 

Updates `WRF_cpl/module_wrf_HYDRO.F` to use the new `MPP_LAND_INIT(global_nx, global_ny)` initializer in mpp_land.F so that coupled model can run without crashing.  Also copies the WRF-supplied MPI_COMM_WORLD to HYDRO_COMM_WORLD, since MPP assumes that MPI has already been initialized and skips this step.

TESTS CONDUCTED: 

Ran WRF-Hydro Front Range Coupled Test Case and output was reasonable (no regression test for this case)

NOTES:

The code compiled originally due to the `in_global_nx` and `in_global_ny` arguments being marked as `optional`, but then caused a crash by being assigned to module variables at runtime. If the parameters can be optional, additional code should be added to handle the not-present case.

### DIFF
--- a/trunk/NDHMS/CPL/WRF_cpl/module_wrf_HYDRO.F
+++ b/trunk/NDHMS/CPL/WRF_cpl/module_wrf_HYDRO.F
@@ -22,10 +22,11 @@
 module module_WRF_HYDRO
 
 #ifdef MPP_LAND
+    use mpi
     use module_mpp_land, only: global_nx, global_ny, decompose_data_real, &
                  write_io_real, my_id, mpp_land_bcast_real1, IO_id, &
                 mpp_land_bcast_real, mpp_land_bcast_int1, mpp_land_init
-    use module_CPL_LAND, only: CPL_LAND_INIT, cpl_outdate
+    use module_CPL_LAND, only: CPL_LAND_INIT, cpl_outdate, HYDRO_COMM_WORLD
     use module_hydro_stop, only: HYDRO_stop
 #endif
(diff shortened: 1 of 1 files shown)


trunk/NDHMS/CPL/WRF_cpl/module_wrf_HYDRO.F
```diff
@@ -98,15 +100,15 @@ subroutine wrf_cpl_HYDRO(HYDRO_dt,grid,its,ite,jts,jte)
...
-           call MPP_LAND_INIT()
-
            !yw nlst_rt(did)%nsoil = config_flags%num_soil_layers
            !nlst_rt(did)%nsoil = model_config_rec%num_metgrid_soil_levels
            nlst(did)%nsoil = grid%num_soil_layers
 
          
 #ifdef MPP_LAND
+           call MPI_COMM_DUP(MPI_COMM_WORLD, HYDRO_COMM_WORLD, ierr)
+           call MPP_LAND_INIT(grid%e_we - grid%s_we - 1, grid%e_sn - grid%s_sn - 1)
```
rcabell, line 110: NOTE: These counts are the velocity points, so they need to be reduced by 1 for the actual surface cell counts.

kafitzgerald (approved):

PR #612, opened 2022-04-01
Backport WRF coupler update to recent MPP_LAND_INIT changes (#602)

kafitzgerald:
Changes:
* duplicate WRF's MPI communicator to HYDRO_COMM_WORLD
* call MPP_LAND_INIT with global grid size from WRF `domain` object

Context:
This is already in the master branch, but would be great to get into a v5.2.x bugfix release.  

I was hoping to get this in prior to the upcoming training, but noticed some odd behavior (i.e. a segfault w/ np=2, but not np=1 or np=4) when I built a version of the coupled system using Docker.  Was trying to troubleshoot that for a bit, but didn't have enough time (and/or understanding of the relevant code / architecture - looks like there's a good bit of duplication w/ stuff in hydro and stuff from WRF, but w/ different implementations and not enough overlap to fully replace?) and it sounds like this problem I ran into is a non-issue w/ Intel + IMPI on Cheyenne.

For the training I think we're going to push ahead w/ the unreleased code, but it'd still be nice to get this in if folks are comfortable with it.  While potentially a little buggy w/ GNU + openMPI, at least it works in most cases.  In an ideal world we'd probably troubleshoot a bit more, but not sure if anyone has time to take a look.

### DIFF
--- a/trunk/NDHMS/CPL/WRF_cpl/module_wrf_HYDRO.F
+++ b/trunk/NDHMS/CPL/WRF_cpl/module_wrf_HYDRO.F
@@ -22,10 +22,11 @@
 module module_WRF_HYDRO
 
 #ifdef MPP_LAND
+    use mpi
     use module_mpp_land, only: global_nx, global_ny, decompose_data_real, &
                  write_io_real, my_id, mpp_land_bcast_real1, IO_id, &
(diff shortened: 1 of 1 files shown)


scrasmussen (approved):
LGTM. I built this with GNU & OpenMPI but didn't run it in coupled mode.

rcabell (approved):

PR #670, opened 2023-01-13
Don't read CROCUS namelist if not HRLDAS-coupled

rcabell:
Quick fix to an issue with WRF Coupling where the configuration manager attempts to read the Crocus namelist from the `namelist.hrldas` file, which doesn't exist when not using a HRLDAS (Noah, NoahMP) coupling.


scrasmussen (approved):
LGTM!

Issue #677, opened 2023-01-25
Classic build system reports successful Make when actually not successful


rcabell:
When using the classic build system, a build will fail but still report "Make was successful"

## Expected Behavior
If the build fails, the message should be "Make NOT succssful"

## Current Behavior
Make always reports successful, even if it wasn't

## Possible Solution
I'm guessing this is most often due to the change to `make; make install` in the compile scripts. The first `make` command fails, but the `$?` variables gets clobbered by the `make install`.  This should be separated into a `make`, with a subsequent check, and then if it succeeded, a call to `make install`.

## Steps to Reproduce (for bugs)
1. Bork something up in the code
2. Compile
3. ???
4. Observe the incorrect message

PR #680, opened 2023-02-09
CMake Build Improvements

### LINKED ISSUE #640: Using gfortran < 10.0 may cause compilation error
When compiling with gfortran 9.4, the following error occurs:

```gfortran: error: unrecognized command line option ‘-fallow-argument-mismatch’; did you mean ‘-Wno-argument-mismatch’?```

## Expected Behavior
Compiling with older versions of gfortran should work correctly.

## Current Behavior
Compiling with older versions of gfortran does not work correctly due to use of newer compiler option

## Possible Solution
Add logic during configuration step (either `./configure` script or CMake) to add that flag only if using gfortran > 10.0

### PR DESCRIPTION
**Pull request by rcabell**
TYPE: bugfix/enhancement

KEYWORDS: CMake, PGI

SOURCE: WRF-Hydro Team @ NCAR

DESCRIPTION OF CHANGES:

* In the classic build system, don't report success if the underlying Make process failed.

* In CMake build system, remove `-fallow-argument-mismatch` from gfortran flags, since this
breaks compatibiilty with older GNU suites (< v10) and not needed anymore with `use netcdf`

* Update the CMake build system to use FIND_MPI() and not override the compiler variables with
static `mpif90` wrapper scripts. This also explicitly links MPI libaries to modules that require
it. 

* Support PGI (> v20) and NVHPC in CMake:
 - added compiler flags from arc/macros.mpp.linux
 - added `use ieee_arithmetic, only: isnan => ieee_is_nan` to CROCUS since PGI doesn't
   suport isnan() intrinsic (this may help on other compilers too)

ISSUES:
```
Fixes #640, #677
```

TESTS CONDUCTED: pending

## Checklist
Merging the PR depends on following checklist being completed. Add `X` between each of the square 
brackets if they are completed in the PR itself. If a bullet is not relevant to you, please comment 
on why below the bullet.

 - [X] Closes issues #640 and #677 (An issue must exist or be created to be closed. The issue describes and documents the problem and general solution, the PR describes the technical details of the solution.) 
 - [X] Backwards compatible
 - [-] Requires new files? If so, how to generate them.
 - [ ] Documentation included

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,18 +1,8 @@
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 3.0)
+cmake_policy(SET CMP0074 NEW)
 
-if (NOT CMAKE_C_COMPILER)
-	set (CMAKE_C_COMPILER "mpicc")
-endif (NOT CMAKE_C_COMPILER)
-if (NOT CMAKE_CXX_COMPILER)
-	set (CMAKE_CXX_COMPILER "mpicxx")
-endif (NOT CMAKE_CXX_COMPILER)
-if (NOT CMAKE_Fortran_COMPILER)
-	set (CMAKE_Fortran_COMPILER "mpif90")
-endif (NOT CMAKE_Fortran_COMPILER)
-
+# set project name and version numbers
 project (WRF_Hydro)
-
-#set version numbers
 set (WRF_Hydro_VERSION_MAJOR 5)
 set (WRF_Hydro_VERSION_MINOR 2)
 set (WRF_Hydro_VERSION_PATCH dev)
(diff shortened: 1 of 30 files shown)


scrasmussen:
These changes look good to me. The only thing I would recommend, if you want to, is to remove the `cmake_minimum_required (VERSION 3.0)` from the top of every `CMakeLists.txt` file other than the one in the top directory. It is my understanding it is only needed once at the top of the project and in the future would make it easier to update the version number.

rcabell:
It would be nice to remove the minimum version from each file -- I did a grep to change it out thinking that it wasn't inherited from the upstream CMakeLists.txt. I'll make that change.

scrasmussen (approved):
Looks good to me! Built successfully with Intel and GNU modules

PR #684, opened 2023-02-23
Update Nudging IO to netCDF NF90 API

### LINKED ISSUE #523: Build fails with gfortran >= 10.0.1

rcabell:
TYPE: update

KEYWORDS: nudging, netcdf

SOURCE: WRF-Hydro Team @ NCAR

DESCRIPTION OF CHANGES: 

There were a few remaining uses of the classing Fortran 77 API to netCDF in `module_nudging.io.F`. This PR migrates them to the NF90 API, which works better with modern versions of gfortran.

ISSUE: 

Fixes #523 Build fails with gfortran >= 10.0.1

TESTS CONDUCTED: Pending

NOTES:

There may be more places still using the NF77 API, but they haven't popped up in builds yet. Primary intent here is to be compatibile with gfortran 10+ without needing `-fallow-argument-mismatch`. Stay tuned...

### DIFF
--- a/src/nudging/module_nudging_io.F
+++ b/src/nudging/module_nudging_io.F
@@ -27,7 +27,6 @@ use module_hydro_stop, only: HYDRO_stop
 
 implicit none
 
-
 !========================
 ! lastObs structure, corresponding to nudgingLastObs.YYYY-mm-dd_HH:MM:ss.nc
 ! How observations from the past are carried forward.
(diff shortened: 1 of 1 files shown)


scrasmussen (dismissed):
Love type checking! LGTM 👍

PR #697, opened 2023-06-12
Alma tweaks

rcabell:
TYPE: choose one of bug fix

KEYWORDS: redhat, build, alma

SOURCE: Ryan @ NCAR

DESCRIPTION OF CHANGES: 

RedHat-style installations of gfortran and gfortran-built system libraries place the resulting `.mod` files into a separate path hierarchy (something like `/usr/lib64/gfortran/modules`) because of the incompatibility of compiled MOD files between compiler vendors and versions. Both the CMake-based `FindNetCDF.cmake` and the legacy WRF-Hydro build system expect to find these .mod files in the standard header/include path. This PR adds support to search these additional directories and/or use `nf-config` to find the needed module files. 

As NCAR is moving to RedHat-compatible OSs on smaller clusters, this will be important to support.

TESTS CONDUCTED: Tested with and without RedHat to check for regressions in build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ set (CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/src/cmake-modules)
 set (NETCDF_F90 "YES")
 set (NETCDF_F77 "YES")
 find_package(NetCDF REQUIRED)
-message("-- NetCDF Include Dir: ${NETCDF_INCLUDES}")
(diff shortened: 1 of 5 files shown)


scrasmussen (approved):
LGTM! 

On a side note, adding to my "wishlist" an item to remove the `include netcdf.h` statements so we can just use use statements and get rid of the NETCDF_F77 requirements. I think that should do the trick

PR #698, opened 2023-06-12
Adding missing NoahMP/Utilities dependency on MPP

rcabell:
TYPE: bugfix

KEYWORDS: bugfix, cmake

SOURCE: Ryan @ NCAR

DESCRIPTION OF CHANGES:

Added `target_link_libraries(noahmp_util PUBLIC hydro_mpp)` to `src/Land_models/NoahMP/Utility_routines/CMakeLists.txt` to indicate dependency on MPP/CPL_WRF.F used by `wrf_abort` (which needs access to the `HYDRO_COMM_WORLD`). This is only an issue when attempting to compile with CMake using more than 6 cores.

TESTS CONDUCTED: Compilation with >6 cores now works with CMake (as well as < 6 cores)

NOTES: This is somewhat of a brute-force approach to this problem, as the routine only needs the COMM_WORLD and not the entirety of MPP. Refactoring `module_cpl_land.mod` into a separate CMake library would make this more lightweight and potentially could speed compilation.

### DIFF
--- a/src/Land_models/NoahMP/Utility_routines/CMakeLists.txt
+++ b/src/Land_models/NoahMP/Utility_routines/CMakeLists.txt
@@ -4,3 +4,5 @@ add_library(noahmp_util STATIC
 	module_model_constants.F
 	module_wrf_utilities.F
 )
+
+target_link_libraries(noahmp_util PUBLIC hydro_mpp)


scrasmussen (approved):
LGTM!

PR #711, opened 2023-09-22
Model MPI enhancements

### LINKED ISSUE #710: MPI Performance on Cray EX computer
Model performance on NCAR's new 'Derecho' Cray EX system is worse than previous 'Cheyenne' SGI system. 

## Expected Behavior
In theory, Derecho should be 20-30% faster than Cheyenne.

## Current Behavior
Initial model testing shows WRF-Hydro on Derecho to be 15% slower on average than the same runs on Cheyenne

## Possible Solution
Communications fabric on Cray appears to favor fewer, larger transfers. Change small point-to-point MPI calls in the model to use larger Collective calls.

## Your Environment
* Version of the code used: v5.3.0
* Operating System and version: SUSE Linux Enterprise Server 15 SP4
* Compiler and version: Cray Compiler Environment
* Other relevant information: NCAR Derecho Supercomputer

rcabell:
TYPE: enchantment

KEYWORDS: MPI, Cray, Derecho

SOURCE: Ryan @NCAR

DESCRIPTION OF CHANGES: 

A number of MPI point-to-point calls in `mpp_land.F` were changed to use Collective communications. This improves performance on the new NCAR Cray system "Derecho".

ISSUE: 
Fixes #710 

TESTS CONDUCTED: 

National Water Model v3.0 CONUS Analysis and Assimilation configuration


src/Routing/module_NWM_io.F
```diff
@@ -5677,11 +5677,11 @@ subroutine output_gw_NWM(domainId,iGrid)
    endif
 
    ! Sync all processes up.
-   if(mppFlag .eq. 1) then
-#ifdef MPP_LAND
-      call mpp_land_sync()
-#endif
-   endif
+!    if(mppFlag .eq. 1) then
```
scrasmussen, line 5680: Do we still want this section commented out? Should it be removed?

src/Routing/module_channel_routing.F
```diff
@@ -528,12 +528,12 @@ subroutine SUBMUSKINGCUNGE(    &
       !TML:Added print statement to test qlos function;
       !comment out to prevent excessive file sizes when running model
       !print*, "qloss,dx,WP,WPk,depth,ChannK,qdc,ql,dt,D", qloss,dx,WP,WPk,depth,ChannK,qdc,ql,dt,D
-      if((qloss*dt)/D > ((ql*dt)/D - C4)) then
-           qloss = ql - C4*(D/dt)
-           if (qloss < 0) then
-              print*, 'WARNING CHANNEL LOSS IS NEGATIVE',qloss
-           endif
-      endif
+      ! if((qloss*dt)/D > ((ql*dt)/D - C4)) then
```
scrasmussen, line 531: Do we still want this section commented out? Should it be removed?

src/Routing/module_channel_routing.F
```diff
@@ -528,12 +528,12 @@ subroutine SUBMUSKINGCUNGE(    &
...
+      !      qloss = ql - C4*(D/dt)
+      !      if (qloss < 0) then
+      !         print*, 'WARNING CHANNEL LOSS IS NEGATIVE',qloss
+      !      endif
+      ! endif
```
scrasmussen, line 536: Do we still want this section commented out? Should it be removed?

rcabell:
I removed additional `mpp_land_sync()` calls from the IO routines, and testing results over CONUS (on Derecho) were the same. Some non-bit-identical results but down in the noise, near or at floating-point epsilon.

src/MPP/mpp_land.F
```diff
@@ -2321,500 +2438,501 @@ end subroutine sum_real8
...
+
+   subroutine updateLake_seq(in,nsize,in0)
+      implicit none
+
+      integer :: nsize
+      real, dimension(:), intent(inout), asynchronous :: in
+      real, dimension(:), intent(in) :: in0
+
+      real, dimension(nsize) :: tmp
+      real, dimension(:), allocatable, asynchronous :: prev
```
scrasmussen, line 2725: Did you find that the asynchronous attribute was needed here? I'm just curious because from what I can gather in [this thread](https://fortran-lang.discourse.group/t/when-to-use-the-asynchronous-attribute-and-why/109/5) it wouldn't make a difference in this case but maybe I'm missing something or interpreting things wrong. I think it's probably fine, just interested if you found that it made a difference.

scrasmussen (approved):
Looks good to me! Tested this on the different Croton configs in serial and parallel and things ran to completion. Well the reach lakes config still breaks but that is an existing Croton issue. Happy for the code cleanup and formatting fixes made :raised_hands:

PR #752, opened 2024-05-21
Don't reset LSM accumulations if `RSTRT_SWC` flag is set to 'no reset'

rcabell:
**TYPE:** bug fix

**KEYWORDS:** NoahMP, accumulations, restarts

**SOURCE:** NCAR Internal

**DESCRIPTION OF CHANGES:**  Previously, if the `RSTRT_SWC` flag was set to 0 in the `hydro.namelist` file, only Hydro accumulations were preserved, but LSM accumulations were still reset.

**TESTS CONDUCTED:**

Ran with and without fix, using Croton test case, with daily restarts. When comparing `LDASOUT` and `RESTART` files, accumulation field increase correctly for the following variables:

```UGDRNOFF, ACCET, SFCRUNOFF, UDRUNOFF, ACCPRCP, ACCECAN, ACCEDIR, ACCETRAN```

as expected, during the initial 24 hours, no differences occur.

### DIFF
--- a/src/CPL/NoahMP_cpl/hrldas_drv_HYDRO.F
+++ b/src/CPL/NoahMP_cpl/hrldas_drv_HYDRO.F
@@ -395,6 +395,14 @@
          iocflag = nlst(did)%io_config_outputs
       end subroutine get_iocflag
 
+      subroutine get_rstrt_swc(did, restart_flag)
+         use config_base, only: nlst
+         implicit none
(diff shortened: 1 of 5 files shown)


scrasmussen (approved):
Looks good to me and I always enjoy formatting changes!

Tests done: built, ran and then ran from a restart, looked at a few accumulated variables.

PR #776, opened 2024-10-23
Restore Reach-based Lakes from v5.1.x

rcabell:
TYPE: bug fix

KEYWORDS: reach, lakes, routing

SOURCE: NCAR

DESCRIPTION OF CHANGES:  support for lakes (reservoirs) in non-UDMP reach-based routing was added in version 5.1.1 but subsequently dropped from subsequent releases. This PR adds that functionality back in.

TESTS CONDUCTED: testing offline, will update in comments if necessary

### DIFF
--- a/src/Routing/module_channel_routing.F90
+++ b/src/Routing/module_channel_routing.F90
@@ -858,11 +858,11 @@ END SUBROUTINE SUBMUSKINGCUNGE
 !            endif !!! order(1) .ne. 1
          end do       !--k links
 
-   #ifdef MPP_LAND
+#ifdef MPP_LAND
          call updateLake_seq(RESHT,nlakes,tmpRESHT)
          call updateLake_seq(QLAKEO,nlakes,tmpQLAKEO)
          call updateLake_seq(QLAKEI,nlakes,tmpQLAKEI)
-   #endif
+#endif
 
 !yw check
 !        gQLINK = 0.0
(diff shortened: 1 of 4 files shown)


src/Routing/module_channel_routing.F90
```diff
@@ -801,29 +831,39 @@ Subroutine drive_CHANNEL(did, latval,lonval,KT, IXRT,JXRT, SUBRTSWCRT, &
...
 
-                   else
+                else
                     print *, "FATAL ERROR: no channel option selected"
                     call hydro_stop("In drive_CHANNEL() - no channel option selected")
                    endif
 !            endif !!! order(1) .ne. 1
          end do       !--k links
 
+   #ifdef MPP_LAND
```
scrasmussen, line 861: These `#ifdef` and `#endif` preprocessor statements are indented and are giving my compiler issues. Removing the leading spaces should fix the issue

scrasmussen (changes requested):
Looks good! Double checked the v5.1.1 code and looks like all the needed changes were brought over but with improved formatting 👍

src/Routing/module_channel_routing.F90
rcabell: Whoops! Not sure how that happened - Intel Fortran didn’t seem to care. Fixed now…

scrasmussen (dismissed):
That fix looks good and now the CI is passing (always comforting!)

scrasmussen (approved):
LGTM!

PR #803, opened 2025-01-08
Allow turning lakes off (`lake_option = 0`) for Gridded

rcabell:
TYPE: bug fix

KEYWORDS: lakes, gridded

SOURCE: NCAR

DESCRIPTION OF CHANGES: 
  
* fixes an issue where `lake_option=0` caused an erroneous error message about a missing `route_lake_f` file by resetting the lake count to 0 after the grid is scanned for channels and lakes

### DIFF
--- a/src/Routing/module_RT.F90
+++ b/src/Routing/module_RT.F90
@@ -567,7 +567,7 @@ subroutine getChanDim(did)
 
    return
 
-endif
+  endif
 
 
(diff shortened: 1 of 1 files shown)


scrasmussen (approved):
Tested with gridded Croton case and was able to confirm bug and then that the fix worked. Nice find!

PR #808, opened 2025-01-24
Correctly handle lake_option=3 / no DA case

rcabell:
TYPE: bug fix

KEYWORDS: reservoir,da

SOURCE: NCAR

DESCRIPTION OF CHANGES: 

If `lake_option` was set to 3 (Reservoir DA) but no Reservoir DA flags were set to `.true.` the internal lake_option was not set to [levelpool/off/passthrough] and a cryptic error resulted. This fix now prints an error message and reverts the value to Level Pool if no DA option is selected.


scrasmussen (approved):
LGTM! Arezoo tested the changes and confirmed they worked.

PR #816, opened 2025-03-13
Revert TYPEL(k) check

rcabell:
Revert TYPEL(k) check from == 2 as in drive_channel_RSL back to the == 1 as needed for traditional routing.

TYPE: bugfix

KEYWORDS: reach routing

SOURCE: NCAR

DESCRIPTION OF CHANGES:

Change `if(TYPEL(k) .ne. 2) then ...` back to `if(TYPEL(k) .ne. 1) then ...`. This had been copied from `drive_channel_RSL()` and doesn't work correctly in traditional channel routing.


scrasmussen (approved):
Glad this fixed the issue, LGTM!

PR #839, opened 2025-05-31
Update CMakeLists.txt to support UFS integration

rcabell:
TYPE: enhancement

KEYWORDS: cmake, ufs

SOURCE: NCAR

DESCRIPTION OF CHANGES: 

 * add `hydro_routing_diversions` and `fort_glob` to `wrfhydro_nuopc_linklibs`
 * check if upstream CMake has defined `netCDF::NetCDF_Fortran` and use it instead of calling `find_package` again


scrasmussen (approved):
Looks good to me! The PR improves the capacity to couple as a submodule.

PR #856, opened 2025-10-01
Make diversion initialization errors non-fatal warnings

rcabell:
TYPE: enhancement

KEYWORDS: diversions

SOURCE: NCAR

DESCRIPTION OF CHANGES: make formerly fatal errors in `init_diversions` warnings that just disable diversion processing for the current run

ISSUE: Missing timeslice files caused model to `error stop` instead of falling back to previous, no-diversions behavior

TESTS CONDUCTED: NCAR Internal


arezoorn (approved):
Tested the code and it works when there is no time slices!

PR #909, opened 2026-07-15
Fix gcc compiler errors in NUOPC cap

danrosen25:
TYPE: bug fix

KEYWORDS: nuopc, gcc

SOURCE: Daniel Rosen - NCAR ESMF

DESCRIPTION OF CHANGES:

Fix the NUOPC cap code to support gnu compilers.
- string list must be the exact same length
- integer format specifier `(Iw)` must include width (e.g. `(I0)` or `(I4)`)

NOTES: NUOPC_CompGet changes have been supported since ESMF v8.0.0 (Oct 12, 2020). This change will not be backwards compatible with ESMF versions prior to 8.0.0.

ISSUE: (issue not opened)

TESTS CONDUCTED: Tested using NASA Land Coupler within a ubuntu container with gcc@13 and mpich@5

### DIFF
--- a/src/CPL/NUOPC_cpl/WRFHydro_NUOPC_Cap.F90
+++ b/src/CPL/NUOPC_cpl/WRFHydro_NUOPC_Cap.F90
@@ -364,23 +364,8 @@ module WRFHydro_NUOPC
     rc = ESMF_SUCCESS
 
     ! Query component for name, verbosity, and diagnostic values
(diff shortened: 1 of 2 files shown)


scrasmussen (approved):
Builds on Derecho with ESMF 8.8.1 and 8.9.1. The cleanup and changes to get it to build look good to me, thanks!

rcabell (approved):

